Issue #10, opened 2019-02-25
Add 20 unit tests


vpetersson:
Create 20 unit tests for the agent.

PR #18, opened 2019-03-12
Tests

slilo:


vpetersson:
Comments: 

* Let's move all test specific packages to `requirements-dev.txt` - no need for them to be installed on regular devices
* Fix the Snyk warnings
* Let's move all tests to 'tests' subfolder(s) to enforce a more clean structure
* I doubt we need this many packages added to `requirements.txt` - let's tighten it up.

slilo:
Closes #10 
Run tests with `pytest --cov=agent`

tests/test_agent.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+def test_get_device_id(cert):
+    with mock.patch(
+            'builtins.open',
+            mock.mock_open(read_data=cert),
+            create=True
+    ):
+        device_id = agent.get_device_id()
+        assert device_id == 'localhost'
```
vpetersson: s/localhost/foobar.d.wott.local/

tests/test_agent.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+def test_generate_cert():  # TODO: parse key and csr
+    cert = agent.generate_cert('localhost')
```
vpetersson: s/localhost/foobar.d.wott.local/

tests/test_agent.py
```diff
@@ -0,0 +1,247 @@
...
+    agent.CERT_PATH = str(tmpdir)
+    agent.CLIENT_CERT_PATH = str(crt_path)
+    agent.CLIENT_KEY_PATH = str(key_path)
+    with mock.patch('builtins.print') as prn:
+        with pytest.raises(json.decoder.JSONDecodeError):
+            hello = agent.say_hello()
+        assert mock.call('Hello failed.') in prn.mock_calls
+
+
+def test_uptime(uptime):
```
vpetersson: I don't fully understand this one. Does uptime always end up being 60s when you mock the interface?

tests/cassettes/test_send_ping.yaml
```diff
@@ -0,0 +1,27 @@
+interactions:
```
vpetersson: This is one of the possible states if you have an invalid certificate. If you however do present a proper certificate and auth, you will receive a 200 (or 201 don't remember).

In short, this is a valid mocked result, but only for a "failed" ping so to speak.

tests/test_agent.py
slilo: It's pre-populated value.
Please check https://github.com/slilo/agent/blob/tests/tests/conftest.py#L160
It passed in the arguments (pytest magic).

tests/cassettes/test_send_ping.yaml
slilo: Updated with proper certificate! Both cases are covered now.

PR #32, opened 2019-03-20
Distr info

### LINKED ISSUE #13: Add distribution information
In order to better understand a fleet, we need to know what operating system/distro a given device is running. `lsb_release -a` is a good example of this:

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Raspbian
Description:    Raspbian GNU/Linux 9.8 (stretch)
Release:        9.8
Codename:       stretch
```

We want to show something similar. However, we need to keep some things in mind:

 * We need to support Ubuntu Core
 * We need to support Balena/Resin

slilo:
Closes #13

### DIFF
--- a/agent/rpi_helper.py
+++ b/agent/rpi_helper.py
@@ -1,3 +1,6 @@
+from sh import lsb_release
+
+
 def detect_raspberry_pi():
     metadata = {
         'is_raspberry_pi': None,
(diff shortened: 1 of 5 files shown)


agent/rpi_helper.py
```diff
@@ -19,3 +22,10 @@ def detect_raspberry_pi():
                 metadata['serial_number'] = line.split()[-1]
 
     return metadata
+
+
+def get_distro():
```
vpetersson: This one is a bit more complicated unfortunately. While this will work when you run it as the python module, it will not work in the Snap or on Balena.

vpetersson: Let's do this:

 * Check if the environment variable `SNAP_INSTANCE_NAME` is set. If so, set the distribution to "Snapd" ([reference](https://docs.snapcraft.io/environment-variables/7983))
 * On Balena/Resin, check for the environment variables `RESIN_APP_NAME` or `BALENA_APP_NAME`. If so, set the 'distribution' to 'Balena'

PR #36, opened 2019-04-04
Count the number of ssh logins by reading system journal.

### LINKED ISSUE #13: Add distribution information

a-martynovich:
Read the system journal, count the number of failed attempts logged by sshd.
Closes #13

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
       - run:
           name: Install Python deps in a venv
           command: |
-            sudo apt-get install python3-systemd
+            sudo apt install libsystemd-dev gcc python3-dev pkg-config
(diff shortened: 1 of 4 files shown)


vpetersson:
I fear by hardcoding the strings for the message, we might miss some environments. However, let's stick with it for now.

PR #45, opened 2019-04-18
Add commandline argument.

### LINKED ISSUE #44: Add arg parser to agent for invoking functions
We have a number of [sub-commands](https://github.com/WoTTsecurity/wott-agent-snap/blob/master/snapcraft.yaml#L58-L73) in our snap. With the move to native deb packaging, we lose these.

As such, the best solution is likely to introduce an arg parser to the client, such that these commands can be invoked.

a-martynovich:
Closes #44

### DIFF
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -1,3 +1,5 @@
+import argparse
+import configparser
 import os
 import requests
 import datetime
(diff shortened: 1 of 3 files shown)


agent/__init__.py
```diff
@@ -355,6 +376,9 @@ def main():
...
 
-
-if __name__ == "__main__":
-    main()
+    print("Writing config...")
+    config = configparser.ConfigParser()
+    config['DEFAULT'] = {'claim_token': crt['claim_token']}
+    with open(INI_PATH, 'w') as configfile:
+        config.write(configfile)
+    os.chmod(INI_PATH, 0o700)
```
vpetersson: Should be 600.

PR #48, opened 2019-04-19
Nicely formatted cmdline help.

### LINKED ISSUE #47: Fix argparse helpers
We need to add helpers to argparse such that the format looks user friendly.

Here's a snippet from the official documentation: 

```
$ python myprogram.py --help
usage: myprogram.py [-h] [--foo FOO]

optional arguments:
 -h, --help  show this help message and exit
 --foo FOO   foo help
```

a-martynovich:
Add debug option.
Fixes #47

### DIFF
--- a/agent/__main__.py
+++ b/agent/__main__.py
@@ -5,22 +5,38 @@ from . import run, get_device_id, get_open_ports, say_hello,\
 
 def main():
     actions = {
-        'whoami': get_device_id,
-        'portscan': get_open_ports,
-        'test-cert': say_hello,
-        'claim-token': get_claim_token,
-        'claim-url': get_claim_url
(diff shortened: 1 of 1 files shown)


a-martynovich:
```
$ python3 -m agent --help
usage: wott-agent [-h] [--debug] [action]

Let's Encrypt for IoT (with more bells and whistles).
When called without arguments, generates device certificate
or renews it if necessary.

positional arguments:
  action      One of the following:
              
              portscan     Print open ports.
              claim-url    Print claim URL.
              claim-token  Print claim token.
              whoami       Print device ID.
              test-cert    Validate device certificate.

optional arguments:
  -h, --help  show this help message and exit
  --debug     Enable debug output.

PR #49, opened 2019-04-20
Send open ports in ping request.

a-martynovich:
Fixes "Ping Failed".

### DIFF
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -2,6 +2,7 @@ import configparser
 import os
 import requests
 import datetime
+import json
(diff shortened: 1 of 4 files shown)


a-martynovich:
`test_send_ping` fails because it performs nmap scan which requires root. I guess I should mock `sh.nmap` or something else.

a-martynovich:
Can't mock `sh.nmap`. Mocking just doesn't work on this function, because sh is (quote) "magical". So instead I have to mock `security_helper.nmap_scan`.

PR #50, opened 2019-04-22
Send netstat info.

### LINKED ISSUE #21: Monitor network connections
In order to determine suspicious network activities, we need to be able to sample the network connections from a given device. In a perfect world, we'd be able to monitor all network connections, but this is unrealistic, so sampling is the next best thing.

A proof of concept illustrating this would be to say run `netstat` every 60 seconds (with a jitter) and record these transactions. We can then log the connections and submit them upstream for processing.

a-martynovich:
Closes #21

### DIFF
--- a/agent/security_helper.py
+++ b/agent/security_helper.py
@@ -53,7 +53,9 @@ def netstat_scan():
     """
     Returns all open inet connections with their addresses and PIDs.
     """
-    lines = netstat(['-pa']).stdout.decode().split('\n')
+    out = netstat(['-pa'])
+    lines_str = out.stdout
+    lines = lines_str.decode().split('\n')
     list_start = 0
     list_end = len(lines)
     for i, l in enumerate(lines):
--- a/agent/security_helper.py
+++ b/agent/security_helper.py
@@ -1,6 +1,8 @@
 import iptc
+import psutil
+import socket
 from xml.etree import ElementTree as ET
-from sh import netstat, nmap
+from sh import nmap
 
 
 def nmap_scan(target):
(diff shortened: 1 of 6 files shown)


agent/security_helper.py
```diff
@@ -33,3 +33,36 @@ def is_firewall_enabled():
...
+        'foreign_address': fields[4],
+        'state': state,
+        'program': program}
+
+
+def netstat_scan():
+    """
+    Returns all open inet connections with their addresses and PIDs.
+    """
+    out = netstat(['-pa'])
```
vpetersson: Can we use some library or built-in for this? I don't like relying on `sh` for parsing. It's too fragile. I had to do it for `nmap` because I didn't find an alternative.

a-martynovich: @vpetersson We can use `psutil` [net_connections](https://psutil.readthedocs.io/en/latest/index.html?highlight=net_connections#psutil.net_connections)

vpetersson: That looks like a much more robust solution.

agent/security_helper.py
```diff
@@ -33,3 +35,18 @@ def is_firewall_enabled():
     filter_table = iptc.Table(iptc.Table.FILTER)
     input_chain = next(filter(lambda c: c.name == 'INPUT', filter_table))
     return not len(input_chain.rules) == 0
+
+
+def netstat_scan():
+    """
+    Returns all open inet connections with their addresses and PIDs.
+    """
+    connections = psutil.net_connections(kind='inet')
```
vpetersson: Shouldn't we user `all` instaed of `inet`. It isn't however clear what the difference is in the manual at a quick glance

a-martynovich: @vpetersson `all` also gathers UNIX sockets. We don't need them, I suppose.

vpetersson: Ah ok - yeah we don't need that for now as we primarily focus on remote attack vectors.

vpetersson:
Thanks!

PR #55, opened 2019-04-24
Send Firewall state

a-martynovich:
Relates to WoTTsecurity/api#107


vpetersson:
@a-martynovich tests are failing here

PR #62, opened 2019-04-30
Update installation as python runtime instruction

rptrchv:


vpetersson:
Looks sensible, @rptrchv. Thanks.

PR #71, opened 2019-05-08
Add iptc.easy from python-iptables

a-martynovich:
Revert the python-iptables dependency back to Stretch version (0.11.0)


rptrchv:
LGTM

PR #78, opened 2019-05-11
Don't use iptc.easy for deleting rules.

### LINKED ISSUE #76: Don't use iptc.easy for port/network blocking
Deleting a rule fails in 50% of my test runs because it can't find a rule to delete.

Use a "non-easy" API for that:
```
import iptc

tbl = iptc.Table('filter')
ch = iptc.Chain(tbl, 'INPUT')
for r in ch.rules:
    for m in r.matches:
        print(m.get_all_parameters())
        if m.comment == 'added by WoTT':
            print('deleting')
            ch.delete_rule(r)
            break
```

a-martynovich:
Fixes #76

### DIFF
--- a/agent/security_helper.py
+++ b/agent/security_helper.py
@@ -134,8 +134,8 @@ def prepare_iptables():
 
 def update_iptables(table, chain, rules):
     """
-    Insert new rules from the supplied list,
-    remove those which are not supplied.
+    Delete all rules marked by WOTT_COMMENT.
+    Then insert new rules from the supplied list.
 
     :param table: table name
     :param chain: chain name
@@ -143,7 +143,7 @@ def update_iptables(table, chain, rules):
     :return: None
     """
     iptc_helper.batch_begin()
-    tbl = iptc.Table('filter')
+    tbl = iptc.Table(table)
     tbl.autocommit = False
     ch = iptc.Chain(tbl, chain)
 
--- a/agent/security_helper.py
+++ b/agent/security_helper.py
@@ -41,12 +41,9 @@ def is_firewall_enabled():
 
 def get_firewall_rules():
     """Get all FILTER table rules"""
-    res = {}
-    for chain_name, chain in iptc_helper.dump_table('filter').items():
-        if chain_name == DROP_CHAIN:
-            continue
-        res[chain_name] = [r for r in chain if r.get('comment') != {'comment': WOTT_COMMENT}]
-    return res
+    table = iptc_helper.dump_table('filter').items()
+    return {chain_name: [rule for rule in chain if rule.get('comment') != {'comment': WOTT_COMMENT}]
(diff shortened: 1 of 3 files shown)


agent/security_helper.py
```diff
@@ -38,7 +41,12 @@ def is_firewall_enabled():
 
 def get_firewall_rules():
     """Get all FILTER table rules"""
-    return iptc_helper.dump_table('filter')
+    res = {}
```
rptrchv: Pls use more meaningful names

agent/security_helper.py
```diff
@@ -53,7 +61,7 @@ def netstat_scan():
         'remote_address': c.raddr,
         'status': c.status if c.type == socket.SOCK_STREAM else None,
         'pid': c.pid
-    } for c in connections]
+    } for c in connections if c.raddr and c.laddr]
```
rptrchv: Probably we shouldn't skip connections with empty local address info. They still can be useful for blocking remote connections

agent/security_helper.py
```diff
@@ -131,21 +139,30 @@ def prepare_iptables():
...
-            iptc_helper.delete_rule(table, chain, r)
+    iptc_helper.batch_begin()
+    tbl = iptc.Table(table)
+    tbl.autocommit = False
+    ch = iptc.Chain(tbl, chain)
+
+    for r in ch.rules:
+        for m in r.matches:
+            if m.comment == WOTT_COMMENT:
+                print('deleting')
```
rptrchv: is it ok leaving `print` statements?

a-martynovich: no, it's not

agent/security_helper.py
a-martynovich: @rptrchv if API can handle it now, I'll revert this

rptrchv: yes, it can

agent/security_helper.py
```diff
@@ -61,7 +61,7 @@ def netstat_scan():
         'remote_address': c.raddr,
         'status': c.status if c.type == socket.SOCK_STREAM else None,
         'pid': c.pid
-    } for c in connections if c.raddr and c.laddr]
+    } for c in connections]
```
rptrchv: you've stripped out too much - the check for `c.raddr` not empty is still required

a-martynovich: @rptrchv But API server now handles this case, doesn't it?
In the future we won't use nmap and instead we'll get open ports from netstat report. This is where raddr will be empty.

rptrchv: > But API server now handles this case, doesn't it?
yes, but there's no need to transmit useless data

a-martynovich: @rptrchv it's not useless, as I've said above.

PR #89, opened 2019-05-20
Service and Periodic execution of ping

### LINKED ISSUE #35: Rewrite Agent to a service because we need the ability to a) serve an API and b) to be able to spawn sub-processes.
A quick look at [this StackOverflow post](https://stackoverflow.com/questions/473620/how-do-you-create-a-daemon-in-python) covers the basics on how to best do this.

a-martynovich:
Closes #35 
The new Executor allows to schedule periodic or one-shot execution of a function in separate process. An arbitrary number of functions can be scheduled, but only MAX_WORKERS processes will be spawned, which is the max number of simultaneously executed functions while others are waiting.

If a function crashes it won't destroy the main process. All function's stdout/stderr is attached to the main process. If a function hangs longer than the specified timeout its process gets killed. All spawned processes are destroyed at main process exit.

A function can be scheduled at any time. It is also possible to stop periodic execution of a single function.

### DIFF
--- a/agent/executor.py
+++ b/agent/executor.py
@@ -37,12 +37,12 @@ class Executor():
                        'daemon': daemon}
         self.process = None
         self.oneshot = interval is None
-        self.shouldStop = False
+        self.should_stop = False
         self.debug = debug
 
     async def start(self):
         """ start calling the process periodically """
-        while not self.shouldStop:
+        while not self.should_stop:
             self.executor.submit(self._submit_unpack_kwargs, self.params)
             if self.oneshot:
                 break
@@ -50,7 +50,7 @@ class Executor():
 
     def stop(self):
         """ terminate running process """
-        self.shouldStop = True
+        self.should_stop = True
         if self.process:
             self.process.terminate()
 
--- a/agent/executor.py
+++ b/agent/executor.py
@@ -6,10 +6,6 @@ from multiprocessing import Queue
 from typing import Callable
 from typing import Dict
 from typing import Any
-import logging
-
-
-L = logging.getLogger()
 
 
 class Executor():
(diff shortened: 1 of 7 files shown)


a-martynovich:
@vpetersson This PR is ready for merge.
A note: if an older version of agent (< 0.1.4.9 ) was installed on the device the newer agent won't start automatically after installation. It will start either after a reboot or if you do `apt purge wott-agent` before installing 0.1.5.0.

vpetersson:
@a-martynovich noted on the older versions - that's fine in this stage. How do we control the granularity of the periodic runs and how are they run right now?

a-martynovich:
@vpetersson
https://github.com/WoTTsecurity/agent/blob/61ccadf354c8a99173f58e41ddd047d90c08de02/agent/__main__.py#L51-L59

Here’s how ping is run right now. Hopefully this code speaks for itself.

vpetersson:
Got it - yep. So i guess what we need to do is to start breaking ping apart and classify how often various operations should run.

agent/executor.py
```diff
@@ -0,0 +1,129 @@
...
+        :param callback_timeout: will be called if the process gets killed on timeout
+        :param daemon:
+        """
+        self.interval = interval
+        self.params = {'func': func, 'fn_args': fargs, "p_kwargs": {},
+                       'timeout': timeout, 'callback_timeout': callback_timeout,
+                       'daemon': daemon}
+        self.process = None
+        self.oneshot = interval is None
+        self.shouldStop = False
```
rptrchv: camelCase?

rptrchv:
@a-martynovich is is possible to test your new code?

a-martynovich:
@rptrchv Manually for now. Proper automated testing for this will take a lot of time to code.

rptrchv:
lgtm

PR #92, opened 2019-05-21
Fetch credentials every 15 minutes and save them to file.

### LINKED ISSUE #90: Receive credentials from the server
*Subtask of WoTTsecurity/api#86*
*Requires #89*

Schedule periodic GET request to `/api/.../creds` every 15 minutes. Write the result to a file.

Credentials are in the form of `[{"name": ..., "key": ..., "value": ...}, ...]`. Agent should write `{key: value, ...}` to a file named `{name}.json` (grouping credentials by `name`).

a-martynovich:
Closes #90.

### DIFF
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -422,6 +422,7 @@ def fetch_creds(debug, dev):
         if debug:
             print("[RECEIVED] Fetch creds: code {}".format(creds_req.status_code))
             print("[RECEIVED] Fetch creds: {}".format(creds_req.content))
+        return
     creds = creds_req.json()
 
     print('Credentials retreived.')
@@ -429,6 +430,7 @@ def fetch_creds(debug, dev):
         print('Creds: {}'.format(creds))
 
     config = configparser.ConfigParser()
+    config['DEFAULT'] = creds
     with open(CREDS_PATH, 'w') as configfile:
         config.write(configfile)
     os.chmod(CREDS_PATH, 0o600)
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -405,7 +405,17 @@ def renew_expired_cert(csr, device_id, debug=False):
     }
 
 
+def setup_endpoints(dev):
+    if dev:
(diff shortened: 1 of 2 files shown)


vpetersson:
@a-martynovich I don't think this is a good implementation, and here's why. 

Let's imagine that you have two set of credentials for two different applications. Let's call them App0 and App1. 

App0 is not super sensitive. It's a 3rd party software and requires a requires an API key. The fleet owner is  using this same API key across the entire fleet of devices and the service as running as `weather-user`.

App1 on the other hand is very sensitive. It's a unique API key that is used for the device to report in data to the back-end. This service is running as the user 'foobar'.

To mitigate the above problem, we need to provide more granularity for permission such that App0 cannot read App1's credentials and vice versa. As such, all credentials cannot reside in the same file.

Instead of the above approach, I suggest that we move to JSON on disk and have each set of credentials being its own file. That way we can regulate access control for the use case above and regulate such that only the weather-user can access the weather-app and 'foobar' the sensitive API key.

Moreover, by using JSON, it's easier for 3rd party services to read in the file.

Yes, this is only true for services not running as root, but it's at least a first layer of defence.

(ping @rptrchv)

a-martynovich:
@vpetersson I thought we were going to provide an API for this, so that only agent has access to creds store, and it checks who's requesting the creds before giving them away.

How do we differentiate between sets of credentials? As per WoTTsecurity/api#86 you said it's simply a fleet-local key-value store. And now we have these sets?

vpetersson:
> @vpetersson I thought we were going to provide an API for this, so that only agent has access to creds store, and it checks who's requesting the creds before giving them away.

Honestly, I've been going back and forth on this. Right now I'm leaning towards writing them out to disk as we would get some basic permission security with minimal complexity.

> How do we differentiate between sets of credentials? As per [WoTTsecurity/api#86](https://github.com/WoTTsecurity/api/issues/86) you said it's simply a fleet-local key-value store. And now we have these sets?

Yes, you're absolutely right. I've been thinking about this for a fair amount of time and I think I simply forgot to update you guys on my thinking.

vpetersson:
Closed in favor of #107.

PR #101, opened 2019-05-27
Docker scripts for an iptables-enabled container which can run agent

a-martynovich:


Dockerfile
```diff
@@ -1,34 +1,20 @@
...
+FROM debian
 WORKDIR /usr/src/app
-ENV PYTHONUNBUFFERED=1
 
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends build-essential libssl-dev libffi-dev libltdl-dev curl && \
-    apt-get clean
+    apt-get install -y build-essential libssl-dev libffi-dev libltdl-dev openssh-server git net-tools \
+        iptables inetutils-ping wget curl pkg-config libsystemd-dev nmap python3-iptables python3-setuptools python3-pip && \
+    service ssh start && \
```
vpetersson: SSH should never be enabled in a docker container. Use `docker exec` instead.

Dockerfile
```diff
@@ -1,34 +1,20 @@
...
-    apt-get clean
-
-RUN git clone https://github.com/square/ghostunnel.git $GOPATH/src/github.com/square/ghostunnel && \
-    cd  $GOPATH/src/github.com/square/ghostunnel && \
-    git checkout -b v1.3.0 && \
-    make
-
-## Runtime container
-FROM python:3.7-slim-stretch
+# docker run --net=host --cap-add=NET_ADMIN --cap-add=SYS_ADMIN  -v /sys/fs/cgroup:/sys/fs/cgroup:ro --security-opt seccomp:unconfined -v ~/Documents/GreatFruit/WoTT:/usr/src/app -Pit mydebian
```
vpetersson: Let's move this to a bash script instead.

Dockerfile
```diff
@@ -1,34 +1,20 @@
...
+FROM debian
```
vpetersson: Let's pin this better to say `debian:stretch`

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+version: '3'
+services:
+  main:
+    command: bash
```
vpetersson: It would make more sense if we launched the wott-agent in foreground mode (i know we don't have this feature yet).

a-martynovich: @vpetersson We do have this feature, but this container is for running agent manually for development, not for keeping it running all the time.

Dockerfile
a-martynovich: I'll remove this, this command is exactly what docker-compose does.

vpetersson: Ah right.

docker-compose.yml
vpetersson: Alright - fair enough.

PR #103, opened 2019-05-28
Fix for piling up of ipv4 rules 

a-martynovich:


a-martynovich:
that still doesn't work

PR #107, opened 2019-05-29
#90 Add grouping by name to fetch_credentials()

### LINKED ISSUE #90: Receive credentials from the server

eugenosm:
Closes #90

### DIFF
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -441,6 +441,10 @@ def fetch_credentials(debug, dev):
     if debug:
         print('Creds: {}'.format(creds))
 
+    for f in os.listdir(os.path.join(CREDS_PATH)):
+        if f.endswith(".json"):
+            os.remove(os.path.join(dir, f))
+
     by_names = {}
     for cred in creds:
         name = cred['name']
@@ -451,9 +455,6 @@ def fetch_credentials(debug, dev):
     for name in by_names:
         creds_fname = os.path.join(CREDS_PATH, "{}.json".format(name))
         stored_creds = {}
-        if os.path.isfile(creds_fname):
-            with open(creds_fname) as json_file:
-                stored_creds = json.load(json_file)
 
         for cred in by_names[name]:
             stored_creds[cred] = by_names[name][cred]
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -443,7 +443,8 @@ def fetch_credentials(debug, dev):
 
     if not os.path.exists(CREDS_PATH):
         os.mkdir(CREDS_PATH, 0x700)
-    elif not os.path.isdir(CREDS_PATH):
+        
+    if not os.path.isdir(CREDS_PATH):
         print("there is file named as our credentials dir({}), that's strange...delete it".format(CREDS_PATH))
         exit(1)
 
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -443,9 +443,9 @@ def fetch_credentials(debug, dev):
 
     if not os.path.exists(CREDS_PATH):
         os.mkdir(CREDS_PATH, 0x700)
-        
+
     if not os.path.isdir(CREDS_PATH):
-        print("there is file named as our credentials dir({}), that's strange...delete it".format(CREDS_PATH))
+        print("there is file named as our credentials dir({}), that's strange...".format(CREDS_PATH))
         exit(1)
 
     for f in os.listdir(os.path.join(CREDS_PATH)):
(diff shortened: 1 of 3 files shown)


agent/__init__.py
```diff
@@ -404,14 +406,68 @@ def renew_expired_cert(csr, device_id, debug=False):
...
+def setup_endpoints(dev):
     if dev:
         global WOTT_ENDPOINT, MTLS_ENDPOINT, DASH_ENDPOINT
         endpoint = os.getenv('WOTT_ENDPOINT', 'http://localhost')
         DASH_ENDPOINT = endpoint + ':' + str(DASH_DEV_PORT)
         WOTT_ENDPOINT = endpoint + ':' + str(WOTT_DEV_PORT) + '/api'
         MTLS_ENDPOINT = endpoint + ':' + str(MTLS_DEV_PORT) + '/api'
 
+
+def fetch_creds(debug, dev):
```
vpetersson: Don't abbreviate words in functions.

agent/__main__.py
```diff
@@ -50,12 +50,19 @@ def main():
 
 PING_INTERVAL = 60 * 60
 PING_TIMEOUT = 10 * 60
+CREDS_INTERVAL = 15 * 60
```
vpetersson: Same as above. Don't abbreviate.

agent/__init__.py
```diff
@@ -44,6 +45,7 @@
 CA_CERT_PATH = os.path.join(CERT_PATH, 'ca.crt')
 COMBINED_PEM_PATH = os.path.join(CERT_PATH, 'combined.pem')
 INI_PATH = os.path.join(CONFIG_PATH, 'config.ini')
+CREDS_PATH = os.path.join(CONFIG_PATH, 'credentials')
```
vpetersson: Same.

agent/__init__.py
```diff
@@ -404,14 +406,68 @@ def renew_expired_cert(csr, device_id, debug=False):
...
+
+    by_names = {}
+    for cred in creds:
+        name = cred['name']
+        if name not in by_names:
+            by_names[name] = {}
+        by_names[name][cred['key']] = cred['value']
+
+    for name in by_names:
+        creds_fname = os.path.join(CREDS_PATH, "{}.json".format(name))
```
vpetersson: This is not a filename, but a file path. Also, don't abbreviate this much as it makes the code harder to read. Readability trumps nifty tricks.

agent/__init__.py
```diff
@@ -404,14 +406,68 @@ def renew_expired_cert(csr, device_id, debug=False):
...
+            print("[RECEIVED] Fetch creds: code {}".format(creds_req.status_code))
+            print("[RECEIVED] Fetch creds: {}".format(creds_req.content))
+        return
+    creds = creds_req.json()
+
+    print('Credentials retreived.')
+    if debug:
+        print('Creds: {}'.format(creds))
+
+    by_names = {}
```
vpetersson: What does 'by_name" mean?

agent/__init__.py
eugenosm: noted

agent/__main__.py
eugenosm: +

agent/__init__.py
eugenosm: +

agent/__init__.py
eugenosm: noted

agent/__init__.py
eugenosm: the credentials was got in request are in array, so to make only one writes for each file resort it to dictionary by their 'name'...so may be not actually right var name

agent/__init__.py
vpetersson: This is still not resolved.

a-martynovich:
@eugenosm I just tested your branch and found out that the folder residing in `CREDS_PATH` is not created if it does not exist. It should be created.

eugenosm:
and what we'll do if there is file with the name of our credentials dir?
cry something and exit? or delete file and create dir?

a-martynovich:
Tests look much better now. Only I would parse JSON [here](https://github.com/WoTTsecurity/agent/pull/107/files#diff-621c77fea2394a9a79831e44a3b482b2R518) and compare dicts instead of byte-strings, but that's splitting hairs.

a-martynovich:
LGTM

PR #113, opened 2019-05-31
Fix default password detection

a-martynovich:
Closes wottsecurity/api#232


vpetersson:
@a-martynovich if we're going down this path, we should probably user pairs (i.e. pi:raspberry) because in other distros, it won't be the user 'pi'.

a-martynovich:
@vpetersson I'll do that after I'm done with other urgent tasks if there still be time left.

vpetersson:
@a-martynovich let's fix the credential pairs before merging this.

a-martynovich:
@vpetersson What do you mean by that? And how does this affect credentials?

vpetersson:
What i mean is that we should check against "pi:raspberry" (i.e a mapping) rather than just the password. In the future, we will have more (such as "root:password") for instance.

PR #117, opened 2019-06-03
Apply iptables changes before gathering and sending them.

### LINKED ISSUE #114: Firewall status not being updated due to execution order
If you make a firewall change, the agent must run twice for the changes to be picked up in the interface. This is likely because we collect firewall status *before* we apply them.

a-martynovich:
Closes #114
Requires wottsecurity/api#246


vpetersson:
@a-martynovich tests failing

Issue #95, opened 2019-05-24
Publish version of agent as part of payload


vpetersson:
For debugging and troubleshooting reasons, we want to know what version a given device is running from the dashboard. Let's submit this as part of the payload and display it under the "Software" tab.

a-martynovich:
took this

PR #118, opened 2019-06-03
Get version from latest git commit.

### LINKED ISSUE #96: Consolidate Python and Deb version
As it stands right now, we have two different versions of the agent (in the same repo):

* The Python version
* The Debian package version

let's consolidate the two are coupled right now anyway.

a-martynovich:
Closes #96 #95


vpetersson:
@a-martynovich interesting. Surely there must be some standard for this, right?

a-martynovich:
@vpetersson There isn't. But `<module>.__version__` is a de-facto standard.

vpetersson:
> @vpetersson There isn't. But `<module>.__version__` is a de-facto standard.

Roger. Let's go for that then. I guess it doesn't solve the release notes fully thought, or does it?

a-martynovich:
@vpetersson there isn’t an easy way for this. Currently only the last commit message is written to the changelog which I thought was enough. The “right” way is to version it according to git tags, but since we don’t have them having full changelog doesn’t make sense either.

vpetersson:
@a-martynovich i think this is fine for now. Are we sure however that the latest version will be used in this scheme? I.e. if we have a.b.c.x and a.b.c.y, where both x and y are git hashes, how does debian know which is the most recent one?

a-martynovich:
@vpetersson The version format is now VERSION.BUILD.HASH where VERSION is what's in VERSION file ("0.1.5" right now), BUILD is CircleCI build number and HASH is a short git hash.
0.1.5.360~aabbcc > 0.1.5.359~bbccdd because 360 > 359.

vpetersson:
Ah ok. That makes sense. Good one.

version.py
```diff
@@ -0,0 +1,34 @@
...
+    build_number = getenv('CIRCLE_BUILD_NUM', '0')
+    return ver, msg, commit, build_number
+
+
+def version_string(ver, msg, commit, n):
+    return '{}.{}~{}'.format(ver, n, commit[:7])
+
+
+def write_changelog():
+    import debian.changelog
```
rptrchv: What's the purpose of importing it here?

a-martynovich: @rptrchv To write debian/changelog. We only need it when building deb package, so we import it here to avoid adding extra dependencies when we don't need them.

rptrchv: I see

PR #121, opened 2019-06-04
 Send NAT and MANGLE tables, v4 and v6.

a-martynovich:
Closes wottsecurity/api#148


agent/iptc_helper.py
```diff
@@ -218,7 +218,9 @@ def test_target(name, value, ipv6=False):
 def get_policy(table, chain, ipv6=False):
     """ Return the default policy of chain in a table """
     iptc_chain = _iptc_getchain(table, chain, ipv6)
-    return iptc_chain.get_policy().name
+    p = iptc_chain.get_policy()
```
vpetersson: @a-martynovich let's use better variable names that actually explains what it is.

vpetersson: Thanks!

PR #122, opened 2019-06-05
Interprocess Locker

### LINKED ISSUE #110: Avoid running multiple same operations at the same time
Running `send_ping` and `fetch_credentials` at the same time is fine; running two `send_ping`s or two `fetch_credentials` is not.

Devise a way to reliably lock a cross-process object and unlock it.

An object could be a "lock file" in `/var/lock` named smth like `wott-agent.ping.lock` or `wott-agent.creds.lock` containing the PID of the process locking it. The actual locking can be done with [fcntl.lockf](https://docs.python.org/3/library/fcntl.html#fcntl.lockf).

There's a [ready to use library](https://github.com/JoshData/exclusiveprocess) for that which can be used, but it needs to be tested first.

*Question: what happens if a process holding the lock dies? How do we recover from that?*

### LINKED ISSUE #124: Version errors
*Subtask of #96*

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.5/json/__init__.py", line 230, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python3.5/json/encoder.py", line 198, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.5/json/encoder.py", line 256, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.5/json/encoder.py", line 179, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: wott-agent 0.1.5.0-7d20ed2 (/usr/lib/python3/dist-packages) is not JSON serializable

pi@raspberrypi:~ $ sudo dpkg -l | grep agent
ii  wott-agent                      0.1.5.284~7d20ed2            all          Let's Encrypt for IoT (with more bells and whistles).

```

* Build number didn't get included into the Python version. Probably because debhelper ignores all env vars and needs them to be specified manually. Also, the tilde character transformed into dash. The deb version is fine.
* The error which caused the traceback can be fixed by simply using `str(__version__)`.

a-martynovich:
Closes #110 
Fixes #124

### DIFF
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -306,7 +306,7 @@ def send_ping(debug=False, dev=False):
         'app_armor_enabled': security_helper.is_app_armor_enabled(),
         'logins': journal_helper.logins_last_hour(),
         'default_password': security_helper.check_for_default_passwords(CONFIG_PATH),
-        'agent_version': __version__
+        'agent_version': str(__version__)
     }
 
     rpi_metadata = rpi_helper.detect_raspberry_pi()
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -474,155 +474,156 @@ def setup_endpoints(dev, debug):
 
 
 def fetch_credentials(debug, dev):
+    with Locker():
+        setup_endpoints(dev, debug)
+        print('Fetching credentials...')
+        can_read_cert()
 
-    setup_endpoints(dev, debug)
-    print('Fetching credentials...')
-    can_read_cert()
+        credentials_req = requests.get(
+            '{}/v0.2/creds'.format(MTLS_ENDPOINT),
+            cert=(CLIENT_CERT_PATH, CLIENT_KEY_PATH),
+            headers={
+                'SSL-CLIENT-SUBJECT-DN': 'CN=' + get_device_id(),
+                'SSL-CLIENT-VERIFY': 'SUCCESS'
+            } if dev else {}
+        )
+        if not credentials_req.ok:
+            print('Fetching failed.')
+            if debug:
+                print("[RECEIVED] Fetch credentials: code {}".format(credentials_req.status_code))
+                print("[RECEIVED] Fetch credentials: {}".format(credentials_req.content))
+            return
+        credentials = credentials_req.json()
 
-    credentials_req = requests.get(
-        '{}/v0.2/creds'.format(MTLS_ENDPOINT),
-        cert=(CLIENT_CERT_PATH, CLIENT_KEY_PATH),
-        headers={
-            'SSL-CLIENT-SUBJECT-DN': 'CN=' + get_device_id(),
-            'SSL-CLIENT-VERIFY': 'SUCCESS'
-        } if dev else {}
-    )
-    if not credentials_req.ok:
-        print('Fetching failed.')
+        print('Credentials retreived.')
         if debug:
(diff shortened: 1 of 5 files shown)


agent/__init__.py
```diff
@@ -465,155 +467,156 @@ def setup_endpoints(dev, debug):
...
-    if not os.path.exists(CREDENTIALS_PATH):
-        os.mkdir(CREDENTIALS_PATH, 0o700)
+        if not os.path.isdir(CREDENTIALS_PATH):
+            print("there is file named as our credentials dir({}), that's strange...".format(CREDENTIALS_PATH))
+            exit(1)
 
-    if not os.path.isdir(CREDENTIALS_PATH):
-        print("there is file named as our credentials dir({}), that's strange...".format(CREDENTIALS_PATH))
-        exit(1)
+        for f in os.listdir(os.path.join(CREDENTIALS_PATH)):
```
vpetersson: Let's try to avoid single character variable names.

agent/executor.py
```diff
@@ -122,8 +123,29 @@ def schedule(executor: Executor) -> asyncio.Future:
...
+    future = asyncio.ensure_future(awaitable)
+    return future
 
 
 def spin():
     asyncio.get_event_loop().run_forever()
+
+
+class Locker:
+    LOCKDIR = '/var/lock'
```
vpetersson: We should probably use a lockfile within our directory, because we can't make assumptions on where we can write on certain platforms (such as within a docker container or in a Snap).

agent/__init__.py
a-martynovich: @vpetersson I didn't write this. It was written in #107

vpetersson:
@a-martynovich don't we run a chance of dead lockfiles if the app crashes midst run before undoing the lock?

a-martynovich:
@vpetersson As soon as the process dies its lock gets released by the kernel automatically.

agent/__init__.py
vpetersson: Sure, but that doesn't mean we can't fix it now. We should always fix issues when we run across them, regardless who the original author said code is.

agent/executor.py
a-martynovich: @vpetersson I was trying to follow https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s09.html .

I'm convinced /var/lock should be available for snaps. I haven't tried it myself. The snap docs say that every snap gets its private `/tmp`, though. But, when was the last time we tried running agent as a snap? Perhaps something else is not allowed, like forking? 

In Docker container `/var/lock` needs to be mounted which I did in the recently merged docker-compose.yml.

Even though our lock file is not shared by other applications it is used by multiple instances of our application. A predefined directory like `/var/lock` already exists when we run our app for the first time. But our directory (like /opt/wott) may or may not exist, depending on how the app was installed. And if it does not exist the app will try to create it, and guess what can happen if two instances try to do that at the same time. So storing the lock in our directory is not a problem, creating the directory is a problem.

If a user installs our app as a deb package, `/opt/wott` will be created at installation. Same for snap. Thus we will only have this non-existing directory issue if our app was installed and run manually on a clean system.

**tl;dr**: I did what you asked for, but not sure if it's right.

vpetersson:
@a-martynovich merge conflict

PR #133, opened 2019-06-13
Dockerfile: install all required dependencies with apt.

a-martynovich:


vpetersson:
Is this really a good idea? We have far less control over the versions. However, playing the devil's advocate on myself, this is pretty much how our deb files are being installed.

a-martynovich:
@vpetersson Python debs rarely get updated, at least on Stretch.
The reason why we need this is because we need to have environment as close to the "real" Raspbian as possible. For example (and this is the main reason why I'm doing this) you won't get a working python-iptables v0.11 on pip. However it is in Stretch repository and it's the latest you can get. Same goes to all other Python deps.

This particular PR installs all required deps because otherwise no deps will be installed at all. @rptrchv noticed this.

vpetersson:
That's a fair point. Accepting and merging.

Issue #125, opened 2019-06-07
Make agent "runtime aware"


vpetersson:
We currently support four different kind of runtime environments:

 * Debian package
 * Python package
 * Docker
 * Ubuntu Snap

Because we cannot do all operations in all these different environments, we need to first be able to detect what environment we are operating in. Once we have done this, we need to disable certain functionalities, such as port scanning and password checks).

We also need to report the runtime environment upstream, just like the version.

a-martynovich:
@vpetersson Debian Package and Python Package are not runtime environments. They are the way the application was installed. Everything else stays the same, nothing changes in the way the application works.

Docker also shouldn't be in this category, because if we're running Debian in Docker it's still Debian, even though it can't have journald/systemd and may have iptables disabled.

Here's what I think it should look like:

- Debian/Ubuntu app
    - Installation:
       - Installed debian package
       - Installed as python package
       - Uninstalled (ran the code directly)
    - Virtualisation:
      - Live
      - Virtual Machine (VirtualBox, VMware, ...)
      - Dockerized
- Ubuntu snap
    - *Virtualization*?

vpetersson:
> Debian Package and Python Package are not runtime environments. 

Yes, technically speaking, you are right. However, there is a reason why I want to group them like this. That reason is tightly coupled with the *capabilities*. For Debian runtime, we can do everything we need to do (manage firewall, check passwords, scan ports etc). 

For Snaps and Docker (i.e. [Balena](https://www.balena.io)), the only thing we can really do is credential management. As such, we need to detect this and hide all security features from the dashboard.

The python library is somewhat of an edge case.

a-martynovich:
One can install a Debian package or a Python package in Debian under Docker. Same goes to the “real” Debian.
In Docker we can manage iptables with CAP_NET_ADMIN. We can also check for passwords (inside the container). But if you’re talking about Balena, then we should only consider Balena in this case.

What I’m saying is, the groups you’re speaking of are not mutually exclusive.

vpetersson:
> One can install a Debian package or a Python package in Debian under Docker. Same goes to the “real” Debian.

Yes and no. What we care about is the operating environment (what I referred to loosely as "runtime environment"). If you have a better word for this, I'm all ears.

> In Docker we can manage iptables with CAP_NET_ADMIN. We can also check for passwords (inside the container). 

Yes, but we don't care about the password inside the docker container because it's unlikely that login services are exposed. 

[...]

PR #134, opened 2019-06-13
Refs #125. Adds confinement detection

vpetersson:
* Detects Docker and Snap environment.
* Makes ping call dependent on environment.

### DIFF
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -36,11 +36,11 @@ DASH_ENDPOINT = WOTT_ENDPOINT.replace('api', 'dash')
 DASH_DEV_PORT = 8000
 WOTT_DEV_PORT = 8001
 MTLS_DEV_PORT = 8002
-CONFINEMENT = rpi_helper.detect_confinement()
+CONFINEMENT = rpi_helper.DetectConfinement
 
 # Conditional handling for if we're running
 # inside a Snap.
-if CONFINEMENT['snap']:
+if CONFINEMENT.SNAP:
     CONFIG_PATH = CERT_PATH = os.getenv('SNAP_DATA')
 else:
     CERT_PATH = os.getenv('CERT_PATH', '/opt/wott/certs')
@@ -299,13 +299,13 @@ def send_ping(debug=False, dev=False):
     }
 
     # Things we can't do within a Snap or Docker
-    if not (CONFINEMENT['snap'] or CONFINEMENT['docker']):
+    if not (CONFINEMENT.SNAP or CONFINEMENT.DOCKER):
         payload['processes'] = security_helper.process_scan(),
         payload['logins'] = journal_helper.logins_last_hour(),
         payload['default_password'] = security_helper.check_for_default_passwords(CONFIG_PATH)
 
     # Things we cannot do in Docker
-    if not CONFINEMENT['docker']:
+    if not CONFINEMENT.DOCKER:
         blocklist = ping.json()
         security_helper.block_ports(blocklist.get('block_ports', {'tcp': [], 'udp': []}))
         security_helper.block_networks(blocklist.get('block_networks', []))
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -300,12 +300,10 @@ def send_ping(debug=False, dev=False):
 
     # Things we can't do within a Snap or Docker
     if not (CONFINEMENT['snap'] or CONFINEMENT['docker']):
-        payload += {
-            'processes': security_helper.process_scan(),
-            'logins': journal_helper.logins_last_hour(),
-            'default_password': security_helper.check_for_default_passwords(
+        payload['processes'] = security_helper.process_scan(),
+        payload['logins'] = journal_helper.logins_last_hour(),
+        payload['default_password'] = security_helper.check_for_default_passwords(
                 CONFIG_PATH),
-        }
 
     # Things we cannot do in Docker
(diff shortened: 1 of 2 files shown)


agent/__init__.py
```diff
@@ -287,28 +288,40 @@ def send_ping(debug=False, dev=False):
...
+                CONFIG_PATH),
+        }
+
+    # Things we cannot do in Docker
+    if not CONFINEMENT['docker']:
+        blocklist = ping.json()
+        security_helper.block_ports(blocklist.get('block_ports', {'tcp': [], 'udp': []}))
+        security_helper.block_networks(blocklist.get('block_networks', []))
+
+        payload += {
```
a-martynovich: `TypeError: unsupported operand type(s) for +=: 'dict' and 'dict'`

agent/rpi_helper.py
```diff
@@ -19,3 +22,31 @@ def detect_raspberry_pi():
...
+    """
+    Returns the confinement environment.
+    This function is used to control features.
+    """
+
+    metadata = {
+        'docker': None,
+        'balena': None,
+        'snap': None,
+    }
```
a-martynovich: This logic will work, but all those fields will either be `None` or `True`. Not quite straight-forward.

agent/rpi_helper.py
```diff
@@ -19,3 +22,31 @@ def detect_raspberry_pi():
...
+
+    # Detect if running inside Docker
+    # Credits: https://stackoverflow.com/a/42674935/346054
+    with open('/proc/1/cgroup', 'rt') as ifh:
+        metadata['docker'] = 'docker' in ifh.read()
+
+    # Detect if running inside Balena
+    if os.getenv('BALENA', False) or os.getenv('RESIN', False):
```
a-martynovich: simply `os.getenv('var')` does the same.

agent/rpi_helper.py
a-martynovich: Also, this asks for an enum. For example the function can return `[Confinement.DOCKER, Confinement.BALENA]` and you can then use it like this:
```
if Confinement.DOCKER in confinement:
...
```

a-martynovich:
In addition to the code comments we should test this code with API server to see if it accepts ping data with most of the fields missing. I'm not sure it does after the recent refactoring.
@rptrchv could you please do this?

rptrchv:
> In addition to the code comments we should test this code with API server to see if it accepts ping data with most of the fields missing. I'm not sure it does after the recent refactoring.
> @rptrchv could you please do this?

@vpetersson @a-martynovich Just finish the agent side of this solution and then I'll apply all required changes to the server side code

agent/__init__.py
vpetersson: Ok, this is just me being stupid. Of course.Fixing.

agent/rpi_helper.py
vpetersson: I've tweaked this  bit per your comment (with False as defaults). `enum` looks like a good approach, but let's stick to this for now unless you disagree?

a-martynovich: I disagree. If we're the ones defining the possible values and we know all possible values we should use enum. It is kind of another static check, helps to avoid future errors, analyzer and intellisense-friendly.
Otherwise this is starting to look like Javascript.

a-martynovich:
@rptrchv I'm asking you to do an end-to-end test of this branch with the master branch API server. We want to know if the server can handle the data before we merge this branch. If it can't we'll hold this back until the required changes are made to the server.
Can you do that?

a-martynovich:
I just did end-to-end test of this branch, the server can handle this data. However even though `firewall_enabled`, `scan_info` and `netstat` fields are missing the Security tab is not hidden (I think we did this a while ago). But this is another task I suppose.

agent/rpi_helper.py
vpetersson: Ok, I've refactored this to use `ENUM`. Let me know if this makes more sense to you.

agent/__init__.py
```diff
@@ -36,10 +36,11 @@
 DASH_DEV_PORT = 8000
 WOTT_DEV_PORT = 8001
 MTLS_DEV_PORT = 8002
+CONFINEMENT = rpi_helper.DetectConfinement
 
 # Conditional handling for if we're running
 # inside a Snap.
-if os.getenv('SNAP_NAME'):
+if CONFINEMENT.SNAP:
```
a-martynovich: this will always be True

agent/rpi_helper.py
a-martynovich: @vpetersson You're using it wrong. You're defining enum members to be True or False. Enum is supposed to be a constant. You compare variables to this constant.

Let me rewrite this chunk if you don't mind.
```
class Confinement(Enum):
    DOCKER = 1
    BALENA = 2
    SNAP = 3

def detect_confinement():
    if os.getenv('SNAP'):
        return Confinement.SNAP
    is_docker = 'docker' in open('/proc/1/cgroup', 'rt').read()
    if is_docker and (os.getenv('BALENA') or os.getenv('RESIN')):
        return Confinement.BALENA
    else:
        return Confinement.DOCKER

### USAGE ###
confinement = detect_confinement()
if confinement is Confinement.SNAP:
# ...
````

vpetersson:
@a-martynovich I'm closing out this PR and let you take over.

Issue #127, opened 2019-06-10
Write simple web app example that uses auth backend.


vpetersson:
We have a [simple webapp example](https://github.com/WoTTsecurity/agent/tree/master/docs/examples/simple-webapp) for mTLS, but we'd need one for auth too.

Let's use a similar structure as the last one but let's use this simple [HTTP Basic Auth](http://flask.pocoo.org/snippets/8/) example for Flask to hook it up with credentials generated by the agent.

PR #140, opened 2019-06-18
Re: #127 - Writing HTTPS Basic Auth example for Simple WebApp

fshmcallister:
- Added basic skeleton using [this](http://flask.pocoo.org/snippets/8/) with username: user and password: pass

- README.txt Introduction sketch out (will be updated)

- Requirements file same as WebApp from mTLS example


fshmcallister:
- Created a new 'app.py' flask application that runs a very simple WebApp requiring HTTP authentication access via a username and password

- Added WoTT credentials to 'app.py' in JSON format to set username and password through WoTT credentials

- Wrote and finessed documentation walking through Basic HTTP auth example with some common errors

- May need to refine curl commands after commit to ensure correct links

docs/examples/simple-webapp-auth/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+**No module named flask**
+```
+File "/Users/user/dir/app_dir/app.py", line 1, in <module>
+    from flask import Flask
+ImportError: No module named flask
+
+```
+
+This can occur if the requirements file does not install properly. A simple fix is installing flask using `pip3 install flask` globally.
```
vpetersson: `pip3 install flask` isn't the right solution here. It's `pip3 install -r requirements.txt`. The source of the problem is likely that the user wasn't in the virtualenv.

fshmcallister: Noted, thank you

PR #143, opened 2019-06-18
Fix revoke interaction with device

rptrchv:
https://github.com/WoTTsecurity/api/issues/261
We don't store 'claimed' and 'claim_token' parameters in local config any longer, instead we always request appropriate values from the server.
The primary purpose of those changes was to make agent work properly after its package reinstallation.

### DIFF
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -208,9 +208,13 @@ def get_ca_cert(debug=False):
 
 def get_claim_token(debug=False, dev=False):
     setup_endpoints(dev, debug)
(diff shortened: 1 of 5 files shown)


agent/__init__.py
```diff
@@ -228,16 +206,23 @@ def get_ca_cert(debug=False):
     return ca.json()['ca_bundle']
 
 
-def get_claim_token():
-    config = configparser.ConfigParser()
-    config.read(INI_PATH)
-    return config['DEFAULT'].get('claim_token', None)
+def get_claim_info(debug=False, dev=False):
+    setup_endpoints(dev, debug)
```
a-martynovich: Need to call `can_read_cert()` before using MTLS to fail fast.

PR #145, opened 2019-06-19
permissions for credentials

### LINKED ISSUE #137: Permissions for credentials
*Needed by Screenly/screenly-ose#1149*
Currently `/opt/wott/credentials` is only accessible by root (`drwx------`) making it inaccessible for other apps such as Screenly.

This directory should be `drwx--x--x`. This means users can access files inside the directory, but can't list or modify them.

Credentials received from the server should have "owner" field (wottsecurity/api#272). Credentials records should be unique by owner-name-key. They should be grouped by owner (before grouping by name) and written to `/opt/wott/credentials/<owner>/<name>.json`. `/opt/wott/credentials/<owner>` should be owned by `<owner>` and have permissions `drwx------`. Files inside it should be owned by `<owner>` and  have permissions have permissions `dr--------`. 

If we had all credentials files under `/opt/wott/credentials` we would expose credentials for the following attack. An attacker tries  `/opt/wott/credentials/notexist.json` and gets "No such file", then tries  `/opt/wott/credentials/exists.json` and gets "Access Denied", meaning that this file exists. Now an attacker knows something he's not supposed to know, such that the file exists.

eugenosm:
closes #137

### DIFF
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -67,7 +67,7 @@ def is_bootstrapping():
     # Create path if it doesn't exist
     if not os.path.isdir(CERT_PATH):
         os.makedirs(CERT_PATH)
-        os.chmod(CERT_PATH, 0o711)
+    os.chmod(CERT_PATH, 0o711)
 
     client_cert = Path(CLIENT_CERT_PATH)
 
@@ -547,7 +547,7 @@ def fetch_credentials(debug, dev):
                     print("There is a file with name of system user in credentials directory ({}).".format(owner_path))
                     exit(1)
                 os.mkdir(owner_path, 0o700)
-                os.chown(owner_path, uid, gid)
+            os.chown(owner_path, uid, gid) # update ownership if user existence in system changed
 
             for name in credentials_grouped[owner]:
                 credential_file_path = os.path.join(owner_path, "{}.json".format(name))
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -472,6 +472,7 @@ def setup_endpoints(dev, debug):
               DASH_ENDPOINT, WOTT_ENDPOINT, MTLS_ENDPOINT
               ))
 
+
 def clear_credentials(dir, debug):
     if debug:
         print("enter: clear_credentials({})".format(dir))
@@ -547,8 +548,7 @@ def fetch_credentials(debug, dev):
             uid = pw.pw_uid
             gid = pw.pw_gid
 
-            owner_path = CREDENTIALS_PATH if not owner \
-                         else os.path.join(CREDENTIALS_PATH, owner)
+            owner_path = CREDENTIALS_PATH if not owner else os.path.join(CREDENTIALS_PATH, owner)
 
             if owner and not os.path.isdir(owner_path):
                 if os.path.exists(owner_path):
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -473,21 +473,21 @@ def setup_endpoints(dev, debug):
               ))
 
 
-def clear_credentials(dir, debug):
-    if debug:
-        print("enter: clear_credentials({})".format(dir))
-    for file in os.listdir(dir):
-        if file == '..' or file == '.':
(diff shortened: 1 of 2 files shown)


eugenosm:
description of changes here:

https://github.com/WoTTsecurity/agent/issues/137

eugenosm:
circleci is very unsatisfied with 
./docs/examples/simple-webapp-auth/app.py

I watch it so most of them are formating and double # in some comments. Is that specified? Needed something to do?

agent/__init__.py
```diff
@@ -506,38 +520,64 @@ def fetch_credentials(debug, dev):
...
 
         if not os.path.isdir(CREDENTIALS_PATH):
             print("there is file named as our credentials dir({}), that's strange...".format(CREDENTIALS_PATH))
             exit(1)
 
-        for file in os.listdir(os.path.join(CREDENTIALS_PATH)):
-            if file.endswith(".json"):
-                os.remove(os.path.join(CREDENTIALS_PATH, file))
+        clear_credentials(CREDENTIALS_PATH)
+        # for file in os.listdir(os.path.join(CREDENTIALS_PATH)):
```
vpetersson: Why is this left in?

agent/__init__.py
```diff
@@ -480,6 +480,20 @@ def setup_endpoints(dev, debug):
 
 
 def fetch_credentials(debug, dev):
+
+    def clear_credentials(path):
```
vpetersson: @a-martynovich Take a close look at this function and imagine how this could be exploited to remove arbitrary files.

agent/__init__.py
```diff
@@ -506,38 +520,64 @@ def fetch_credentials(debug, dev):
...
-            if debug:
-                print('Store credentials: to {} \n {}'.format(credential_file_path, file_credentials))
+            uid = pw.pw_uid
+            gid = pw.pw_gid
+
+            owner_path = CREDENTIALS_PATH if not owner else os.path.join(CREDENTIALS_PATH, owner)
+
+            if owner and not os.path.isdir(owner_path):
+                if os.path.exists(owner_path):
+                    print("there is file named as owner in credentials dir({}), that's strange...".format(owner_path))
```
vpetersson: Please format sentences properly.

a-martynovich:
> circleci is very unsatisfied with
> ./docs/examples/simple-webapp-auth/app.py
> 
> I watch it so most of them are formating and double # in some comments. Is that specified? Needed something to do?

@eugenosm you can run tests locally, just `pytest tests` (if running in a container make sure to run as user `nobody`). I ran them and got errors in `test_fetch_credentials` and `test_fetch_credentials_no_dir`. You need to mock all `os` and `pwd` functions (chmod, getpwnam) because CircleCI tests are not running as root I believe.

agent/__init__.py
eugenosm: oops...now removed

agent/__init__.py
eugenosm: yes...about a glob i forgot. thank you.

agent/__init__.py
eugenosm: changed. i hope correctly.

agent/__init__.py
```diff
@@ -66,7 +67,7 @@ def is_bootstrapping():
     # Create path if it doesn't exist
     if not os.path.isdir(CERT_PATH):
         os.makedirs(CERT_PATH)
-        os.chmod(CERT_PATH, 0o700)
+        os.chmod(CERT_PATH, 0o711)
```
a-martynovich: I think we should chmod these directories unconditionally. Because if we upgrade the older agent these directories will be present, but their mode will be 700.

eugenosm: agreed

a-martynovich:
@eugenosm Please make all printing conditional (if debug flag is set). Currently not only credential names, but also their contents are printed. They contain sensitive data which will be logged to system log and will be visible to everyone.

PR #171, opened 2019-06-24
Adding documentation for Screenly OSE Use case

fshmcallister:
- Includes full README.md with image examples
- Uses WoTT to authenticate Screenly login


docs/examples/screenly/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+Screenly by default allows anyone within the network with the management page IP address to access it and send content. However, it does also provide HTTP authentication- and we can use WoTT's credentials to secure our Screenly device so that we can restrict and verify those who have access to it. 
+
+Screenly offers a free OSE version that you can use. For this example you will need a Raspberry Pi, a monitor for Screenly to project onto; and at least one browser device. You will also need to have an account for the [WoTT Dashboard](https://dash.wott.io) in order to manage WoTT credentials.
+
+## Installing WoTT agent on Screenly OSE
+
+First you will need a Screenly device if you do not already.
+Follow the instructions to install Screenly OSE (the free version) on your Raspberry Pi [here](https://www.screenly.io/ose/). We reccommend you follow the first option and use something like Etcher to flash the SD card with the Screenly boot.
```
vpetersson: "Screenly boot" -> "The Screenly OSE disk image"

docs/examples/screenly/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+Screenly by default allows anyone within the network with the management page IP address to access it and send content. However, it does also provide HTTP authentication- and we can use WoTT's credentials to secure our Screenly device so that we can restrict and verify those who have access to it. 
+
+Screenly offers a free OSE version that you can use. For this example you will need a Raspberry Pi, a monitor for Screenly to project onto; and at least one browser device. You will also need to have an account for the [WoTT Dashboard](https://dash.wott.io) in order to manage WoTT credentials.
+
+## Installing WoTT agent on Screenly OSE
+
+First you will need a Screenly device if you do not already.
+Follow the instructions to install Screenly OSE (the free version) on your Raspberry Pi [here](https://www.screenly.io/ose/). We reccommend you follow the first option and use something like Etcher to flash the SD card with the Screenly boot. 
+**Note** Screenly will overwrite your OS.
```
vpetersson: Screenly -> Screenly OSE

docs/examples/screenly/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+When the Raspberry Pi reboots, it will take you to the Screenly network configuration. You will need to access the SSID as you would a wifi network and enter the password shown. 
+
+![screenly](https://github.com/WoTTsecurity/agent/blob/master/docs/examples/screenly/screenly-setup.jpg)
+
+Then navigate to the [Address](https://www.screenly.io/wifi) shown and login with your network details. This will grant you access to the Screenly management page.
```
vpetersson: Probably not worth writing out the address here as it's better to have them following the interface.

docs/examples/screenly/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+When the Raspberry Pi reboots, it will take you to the Screenly network configuration. You will need to access the SSID as you would a wifi network and enter the password shown. 
+
+![screenly](https://github.com/WoTTsecurity/agent/blob/master/docs/examples/screenly/screenly-setup.jpg)
+
+Then navigate to the [Address](https://www.screenly.io/wifi) shown and login with your network details. This will grant you access to the Screenly management page. 
+
+Now we need to secure access to your Screenly device. If you haven't installed the WoTT Agent already, install [it](https://github.com/WoTTsecurity/agent).
+
+To do this on the Screenly device, you need to access the terminal through `CTRL` + `ALT` + `F1`. To return back to the GUI, it is `CTRL` + `ALT` + `F2`.
```
vpetersson: I've added it to the installer, so they can actually just do `./screenly/.bin/run-upgrade.sh` and select to install the WoTT agent.

docs/examples/screenly/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+![tag](https://github.com/WoTTsecurity/agent/blob/master/docs/examples/screenly/tag.png)
+
+Now navigate to credentials and add a new credential with the following details:
+
+```
+Name = screenly
+Key = login
+Value = username:password
+Owner = user
```
vpetersson: `Owner = pi`

docs/examples/screenly/README.md
```diff
@@ -0,0 +1,130 @@
...
+Tags = screenly-pi
+```
+Where the Owner is a valid Linux username, specifically the Linux user running Screenly (so by default on a Raspberry Pi, `pi`) and Key must be `login` (or the credentials won't be read). Value is the actual username and password of the login denoted by the single field `username:password`. You can change this value to match your own criteria, but it's fine for this example to leave it as is. 
+
+You can change the Name value of `screenly` if you wish, but note that this is how WoTT will call the credential information, so if you do change this name, be aware to use the correct credential name later on. Make sure as well that the Tags match whatever Tag you assigned the Pi earlier.
+
+You now need to set up the screenly configuration file to call the WoTT credentials. This file is in a hidden folder, so to access it do the following:
```
vpetersson: You can skip the text edit part as I've already added that in upstream.

docs/examples/screenly/README.md
```diff
@@ -0,0 +1,130 @@
...
+$ service wott-agent restart
+$ service screenly-web restart
+```
+
+There will now be a file on your Pi `screenly.json`. 
+
+You may need to wait a few seconds for the server to restart. You can check the file exists by running the following command (if your user is `pi`): 
+
+``` 
+$ sudo find /opt/wott/credentials/pi
```
vpetersson: No need for `sudo` here. Also, let's do `cat /opt/wott/credentials/pi/screenly.json` instead as it will catch any error.

docs/examples/screenly/README.md
```diff
@@ -0,0 +1,130 @@
...
+and `ESC` + `:wq` to save.
+
+Your credentials are now all set up. To download them onto the device, you will need to restart the WoTT Agent and the Screenly server:
+
+```
+$ service wott-agent restart
+$ service screenly-web restart
+```
+
+There will now be a file on your Pi `screenly.json`.
```
vpetersson: It might take a few minutes for the file to appear (in particular on older Pis).

docs/examples/screenly/README.md
```diff
@@ -0,0 +1,130 @@
...
+![Chrome](https://github.com/WoTTsecurity/agent/blob/master/docs/examples/screenly/screenly-chrome.png)
+
+If your credentials are correct, you will be successfully logged into the main Screenly management page and greeted by the Schedule Overview. 
+
+## Closing Notes
+
+You can change the WoTT credentials as you wish, but note that data is fetched by `wott-agent service` every 15 minutes, so for access to any immediate changes you implement you will need to restart the WoTT agent and server as before using
+
+``` 
+$ service wott-agent restart
```
vpetersson: You need `sudo` here.

docs/examples/screenly/README.md
fshmcallister: does this assume that the name of the file is 'screenly' ?

docs/examples/screenly/README.md
fshmcallister: I've done owner = user here then specified why in the comments in case the the pi isn't actually named pi. If it's not clear enough, I can change it

docs/examples/screenly/README.md
vpetersson: yep.

docs/examples/screenly/README.md
vpetersson: In Screenly it's always 'pi'. It's hardcoded in many places.

docs/examples/screenly/README.md
fshmcallister: Added a note at the end of the document - because I don't run as root, I personally need to use sudo to access these files

docs/examples/screenly/README.md
```diff
@@ -125,6 +125,6 @@ You can change the WoTT credentials as you wish, but note that data is fetched b
 $ service wott-agent restart
 $ service screenly-web restart
 ```
-Additionally, you can use `pkill -f server.py` to kill the Screenly server, but this will take a few seconds to restart.
+Additionally, you can use `pkill -f server.py` to kill the Screenly OSE server, but this will take a few seconds to restart.
```
vpetersson: `service screenly-web restart` is the same as `pkill -f server.py`.

PR #172, opened 2019-06-24
Fix errors when running/building Snap 

a-martynovich:
wottsecurity/wott-agent-snap#4
Use location mapping in core18.
Skip git commit if can't import git.
If running as snap get the version from snap package.

### DIFF
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -54,7 +54,6 @@ CONFIG_PATH = os.getenv('CONFIG_PATH', '/opt/wott')
 if not os.path.isdir(CONFIG_PATH):
     os.makedirs(CONFIG_PATH)
     os.chmod(CONFIG_PATH, 0o711)
-Locker.LOCKDIR = CONFIG_PATH
 
 # This needs to be adjusted once we have
 # changed the certificate life span from 7 days.
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -46,10 +46,15 @@ WOTT_DEV_PORT = 8001
 MTLS_DEV_PORT = 8002
 CONFINEMENT = detect_confinement()
 
-# Conditional handling for if we're running
-# inside a Snap.
-CERT_PATH = os.getenv('CERT_PATH', '/opt/wott/certs')
 CONFIG_PATH = os.getenv('CONFIG_PATH', '/opt/wott')
+CERT_PATH = os.getenv('CERT_PATH', os.path.join(CONFIG_PATH, 'certs'))
+CREDENTIALS_PATH = os.getenv('CREDENTIALS_PATH', os.path.join(CONFIG_PATH, 'credentials'))
+
+CLIENT_CERT_PATH = os.path.join(CERT_PATH, 'client.crt')
+CLIENT_KEY_PATH = os.path.join(CERT_PATH, 'client.key')
+CA_CERT_PATH = os.path.join(CERT_PATH, 'ca.crt')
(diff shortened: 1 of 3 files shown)


agent/__init__.py
```diff
@@ -42,11 +47,8 @@
 
 # Conditional handling for if we're running
 # inside a Snap.
-if CONFINEMENT == Confinement.SNAP:
-    CONFIG_PATH = CERT_PATH = os.getenv('SNAP_DATA')
-else:
-    CERT_PATH = os.getenv('CERT_PATH', '/opt/wott/certs')
-    CONFIG_PATH = os.getenv('CONFIG_PATH', '/opt/wott')
+CERT_PATH = os.getenv('CERT_PATH', '/opt/wott/certs')
```
vpetersson: Is this path really writeable inside the snap?

a-martynovich: Yes, because it's mapped. See WoTTsecurity/wott-agent-snap#4

vpetersson: I don't think this is the right way of doing it. I think it needs to be in `SNAP_DATA` in order for the Snap to share it with other Snaps using Plugs.

a-martynovich: If so, it needs to be in `SNAP_COMMON` and have a `content` interface declared.
This code doesn't need to be changed, we can change the mapping in snapcraft.yaml.

vpetersson: So everything that we might need to share (such as keys, certs and credentials) need to be in a storage that can be shared. Lockfile etc will never be used by another app, so we can put them elsewhere. 

Also, keep in mind that in a perfect world, we'll move the keys to a TPM, but that's another story.

a-martynovich: > So everything that we might need to share (such as keys, ...

Don't you think we shouldn't share keys?

> Lockfile etc will never be used by another app, so we can put them elsewhere.

That's why I was initially storing lockfiles in /var/lock, but you told me otherwise. Now that I can test snaps myself I'll put them in /var/lock again and see if it works out.

vpetersson: Sorry for the misunderstanding.

version.py
```diff
@@ -1,21 +1,24 @@
...
 
 def version():
     ver = open('VERSION').read().strip()
-    repo = git.Repo('.')
-    head = repo.head.object
-    msg = head.message
-    commit = str(head)
+    try:
+        import git
+        repo = git.Repo('.')
```
rptrchv: pls don't put to the `try` clause anything except the code you suspect to throw an exception

version.py
```diff
@@ -1,21 +1,24 @@
...
+        head = repo.head.object
+        msg = head.message
+        commit = str(head)
+    except ImportError:
+        commit = msg = None
     build_number = getenv('CIRCLE_BUILD_NUM', '0')
     return ver, msg, commit, build_number
 
 
 def version_string(ver, msg, commit, n):
```
rptrchv: a docstring would be very useful here

version.py
a-martynovich: I suspect this code to throw `ImportError`. If it does the rest of the code in this `try` section won't be able to execute. Or do you want me to leave just `import git` there? Are you sure you understand the try-catch semantics?

rptrchv: > Or do you want me to leave just import git there?

yes
> Are you sure you understand the try-catch semantics?

Снимок экрана 2019-06-26 в 16 22 18

a-martynovich: Ok I'm sorry, it's my fault, you're absolutely right.

PR #174, opened 2019-06-25
add device-specific metadata fetching

eugenosm:
closes  https://github.com/WoTTsecurity/api/issues/273

(agent side)

### DIFF
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -240,7 +240,6 @@ def mtls_request(method, url, debug=False, dev=False, requester_name=None, debug
     MTLS  Request.request wrapper function.
     :param method = 'get,'put,'post','delete','patch','head','options'
     :param url: request url string (without endpoint)
-    :param data: json data, None by default
     :param debug: if true then log error messages
     :param dev: if true use dev endpoint and dev headers
     :param requester_name: displayed requester id for error messages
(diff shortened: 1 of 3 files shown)


agent/__init__.py
```diff
@@ -490,6 +491,43 @@ def setup_endpoints(dev, debug):
...
+            if debug:
+                print("[RECEIVED] Fetch device metadata: code {}".format(dev_md_req.status_code))
+                print("[RECEIVED] Fetch device metadata: {}".format(dev_md_req.content))
+            return
+        metadata = dev_md_req.json()
+
+        print('metadata retrieved.')
+
+        if os.path.exists(SECRET_DEV_METADATA_PATH) and not os.path.isfile(SECRET_DEV_METADATA_PATH):
+            print("Error: The filesystem object '{}' is not a file. Looks like a break-in attempt.".printf(
```
a-martynovich: `printf`?

eugenosm: habit from c++/go

vpetersson:
@eugenosm can this be merged before the API changes?

eugenosm:
i'm not sure. it will try to fetch metadata each 15 min in daemon mode.

vpetersson:
> i'm not sure. it will try to fetch metadata each 15 min in daemon mode.

But does that mean it will crash if it cannot get the metadata? If so, i need to hold this.

a-martynovich:
Daemon never crashes even if its children crash

PR #178, opened 2019-06-26
fix credentials permissions

### LINKED ISSUE #177: credentials folder permissions not updated if folder exists
```python
if not os.path.exists(CREDENTIALS_PATH):
            os.mkdir(CREDENTIALS_PATH, 0o711)
```
should be

```python
if not os.path.exists(CREDENTIALS_PATH):
            os.mkdir(CREDENTIALS_PATH, 0o711)
os.chmod(CREDENTIALS_PATH, 0o711)
```

eugenosm:
closes #177

### DIFF
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -524,6 +524,7 @@ def fetch_credentials(debug, dev):
 
         if not os.path.exists(CREDENTIALS_PATH):
             os.mkdir(CREDENTIALS_PATH, 0o711)
+        os.chmod(CREDENTIALS_PATH, 0o711)
 
         if not os.path.isdir(CREDENTIALS_PATH):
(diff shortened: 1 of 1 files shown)


agent/__init__.py
```diff
@@ -524,6 +524,7 @@ def clear_credentials(path):
 
         if not os.path.exists(CREDENTIALS_PATH):
             os.mkdir(CREDENTIALS_PATH, 0o711)
+        os.chmod(CREDENTIALS_PATH, 0o711)
```
vpetersson: Shouldn't this be an 'else' here? kinda redundant otherwise.

eugenosm: it would have any sense only once.

and this issue happens only cause upgrade

PR #179, opened 2019-06-26
refactor mtls requests

### LINKED ISSUE #162: Refactor MTLS requests
*Subtask of #152*

Currently we have
```
headers={
    'SSL-CLIENT-SUBJECT-DN': 'CN=' + get_device_id(),
    'SSL-CLIENT-VERIFY': 'SUCCESS'} if dev else {}
}
```
added to almost all MTLS requests. 

For all such requests use a function which makes requests with these SSL headers (if dev flag is set) and logs errors (if debug flag is set).

eugenosm:
closes #162

### DIFF
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -233,7 +233,8 @@ def mtls_req_error_log(request_url, req_type, requester_name, response):
         requester_name, req_type, response.content))
 
 
-def mtls_request(method, url, debug=False, dev=False, requester_name=None, debug_on_ok=False, **kwargs):
+def mtls_request(method, url, debug=False, dev=False, requester_name=None, debug_on_ok=False,
+                 return_exception=False, **kwargs):
     """
     MTLS  Request.request wrapper function.
     :param method = 'get,'put,'post','delete','patch','head','options'
@@ -243,6 +244,7 @@ def mtls_request(method, url, debug=False, dev=False, requester_name=None, debug
     :param dev: if true use dev endpoint and dev headers
     :param requester_name: displayed requester id for error messages
     :param debug_on_ok: if true with debug then log successful response status/content too
+    :param return_exception if true, then returns tuple ( response, None ) or (None, RequestException)
     :return: response or None (if there was exception raised), or tuple (see above, return_exception)
     """
     try:
@@ -256,11 +258,17 @@ def mtls_request(method, url, debug=False, dev=False, requester_name=None, debug
 
         if (debug_on_ok or not r.ok) and debug:
             mtls_req_error_log(url, method.upper(), requester_name, r)
-        return r
+        if return_exception:
+            return r, None
+        else:
+            return r
 
     except requests.exceptions.RequestException as e:
         print('wott-agent :: mtls_request :: rises exception: {}'.format(e))
-        return None
+        if return_exception:
+            return None, e
+        else:
+            return None
 
 
 def get_claim_token(debug=False, dev=False):
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -211,9 +211,9 @@ def get_ca_cert(debug=False):
 
 def get_mtls_header(debug=False, dev=False):
     return {
-            'SSL-CLIENT-SUBJECT-DN': 'CN=' + get_device_id(debug=debug, dev=dev),
-            'SSL-CLIENT-VERIFY': 'SUCCESS'
-        } if dev else {}
+        'SSL-CLIENT-SUBJECT-DN': 'CN=' + get_device_id(debug=debug, dev=dev),
+        'SSL-CLIENT-VERIFY': 'SUCCESS'
+    } if dev else {}
 
 
 def mtls_req_error_log(request_url, type, requester_name, response):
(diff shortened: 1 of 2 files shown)


eugenosm:
there is also needed to modify fetch_device_metadata() from 
#174  after it would be merged

agent/__init__.py
```diff
@@ -209,28 +209,102 @@ def get_ca_cert(debug=False):
...
+    """
+    if not requester_name:
+        requester_name = "({})".format(request_url)
+    print("wott-agent :: mtls_post_request :: [RECEIVED] {} {}: {}".format(
+        requester_name, req_type, response.status_code))
+    print("wott-agent :: mtls_post_request :: [RECEIVED] {} {}: {}".format(
+        requester_name, req_type, response.content))
+
+
+def mtls_post_request(req, json=None, debug=False, dev=False, requester_name=None, debug_on_ok=False):
```
rptrchv: Снимок экрана 2019-06-27 в 15 20 34

agent/__init__.py
```diff
@@ -209,28 +209,102 @@ def get_ca_cert(debug=False):
...
     can_read_cert()
-    try:
-        response = requests.get('{}/v0.2/claimed'.format(MTLS_ENDPOINT), cert=(CLIENT_CERT_PATH, CLIENT_KEY_PATH),
-                                headers={'SSL-CLIENT-SUBJECT-DN': 'CN=' + get_device_id(),
-                                         'SSL-CLIENT-VERIFY': 'SUCCESS'} if dev else {})
-    except requests.exceptions.ConnectionError:
+
+    response = mtls_get_request('/v0.2/claimed', debug=debug, dev=dev,
+                                requester_name="Get Device Claim Info")
+    if not response or not response.ok:
```
rptrchv: not `if not response`, but `if not response is None`

a-martynovich:
I believe `mtls_get_request()` and `mtls_post_request()` should be merged into one `mtls_request(method, ...)`. Otherwise there's still code duplication.
Also `mtls_post_request` doesn't have `debug_on_ok` option which logs response. But it should have.

And finally, every endpoint is currently supplied as `/v0.2/xxx`. Given the current server implementation it's impossible to have smth like `/v0.2` and `/v0.3` endpoints on the same server, which means all endpoints will have the same prefix. Why not let `mtls_request()` add this prefix to the endpoint? And make sure that endpoint doesn't start with a slash (`mtls_request()` should add a slash).

agent/__init__.py

agent/__init__.py
eugenosm: this was  just cause of:
image

but, i  agree with you. Fixed

eugenosm:
> I believe mtls_get_request() and mtls_post_request() should be merged into one mtls_request(method, ...). Otherwise there's still code duplication.

DRY is a good principle, but all should be in moderation. There is many kind of requests: `get`, `post`, `delete`, `put`, `patch`, e.t.c.   Each have it's function in request, with their own arguments set.  The `mtls_xxx_request()` is just a wrappers, and making a Frankenstain monster is a bad idea.

> Also mtls_post_request doesn't have debug_on_ok option which logs response. But it should have.

???
image

Or you mean when called? But there it also had
image

> And finally, every endpoint is currently supplied as /v0.2/xxx. Given the current server implementation it's impossible to have smth like /v0.2 and /v0.3 endpoints on the same server, which means all endpoints will have the same prefix. Why not let mtls_request() add this prefix to the endpoint? And make sure that endpoint doesn't start with a slash (mtls_request() should add a slash).

Here i agree with you. Fixed.

agent/__init__.py
```diff
@@ -209,28 +209,102 @@ def get_ca_cert(debug=False):
...
+            '{}/v0.2/{}'.format(MTLS_ENDPOINT, req),
+            cert=(CLIENT_CERT_PATH, CLIENT_KEY_PATH),
+            headers=get_mtls_header(debug=debug, dev=dev)
+        )
+
+        if (debug_on_ok or not r.ok) and debug:
+            mtls_req_error_log(req, 'GET', requester_name, r)
+        return r
+
+    except Exception as e:
```
a-martynovich: Swallowing all exceptions is not a good idea. I wouldn't do it at all here.

a-martynovich:
> There is many kind of requests: get, post, delete, put, patch, e.t.c. Each have it's function in request, with their own arguments set. The mtls_xxx_request() is just a wrappers, and making a Frankenstain monster is a bad idea.

We're trying to avoid code duplication here. You have two very similar functions which share a lot of code.
Do it however you want, leave the two functions for get and post. But please try to reduce duplicated code. This was the whole point of the task.

agent/__init__.py
```diff
@@ -209,28 +209,102 @@ def get_ca_cert(debug=False):
...
+    :param req_type: 'GET', 'POST', ...
+    :param requester_name: requester id for message, if None then request_url string used
+    :param response - request response
+    :return: None
+    """
+    if not requester_name:
+        requester_name = "({})".format(request_url)
+    print("wott-agent :: mtls_post_request :: [RECEIVED] {} {}: {}".format(
+        requester_name, req_type, response.status_code))
+    print("wott-agent :: mtls_post_request :: [RECEIVED] {} {}: {}".format(
```
a-martynovich: This will always log `wott-agent :: mtls_post_request ...`, even for `mtls_get_request`.

agent/__init__.py
eugenosm: In any network problems agent clogs the screen with exception and hungs if not in daemon mode.  It not looks good. The only one function where it was handled some way is `get_claim_token()` but that handle only `connection error` exception and usethe same message for bad status and exception.  So i chnage whole exceptions to only requests one and left the way to get it as second result for future use.

eugenosm: moreover we plan to refactor logging from print to logging. then it would be better to remove that exception return of mine and logs unhandled exceptions as errors

PR #188, opened 2019-07-12
automatically claim device by enrollment token in config

eugenosm:
closes: https://github.com/WoTTsecurity/api/issues/304

api side PR here: https://github.com/WoTTsecurity/api/pull/316

### DIFF
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -680,7 +680,6 @@ def enroll_device(enroll_token, claim_token, device_id):
         logger.exception("enroll_device :: rises exception:")
 
 
-
 def run(ping=True, dev=False, logger=logger):
 
     with Locker('ping'):
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -274,6 +274,36 @@ def mtls_request(method, url, dev=False, requester_name=None, log_on_ok=False, r
             return None
 
 
+def try_enroll_in_operation_mode(device_id, dev):
+    enroll_token = get_enroll_token()
+    if enroll_token is None:
+        return
+    logger.info("Enroll token found. Trying to automatically enroll the device.")
+
+    setup_endpoints(dev)
+    response = mtls_request('get', 'claimed', dev=dev, requester_name="Get Device Claim Info")
+    if response is None or not response.ok:
+        logger.error('Did not manage to get claim info from the server.')
+        return
+    logger.debug("[RECEIVED] Get Device Claim Info: {}".format(response))
+    claim_info = response.json()
+    if claim_info['claimed']:
+        logger.info('The device is already claimed. No enrolling required.')
+    else:
+        claim_token = claim_info['claim_token']
+        if not enroll_device(enroll_token, claim_token, device_id):
+            logger.error('Device enrolling failed. Will try next time.')
+            return
+
+    logger.info("Update config...")
+    config = configparser.ConfigParser()
+    config.read(INI_PATH)
+    config.remove_option('DEFAULT', 'enroll_token')
+    with open(INI_PATH, 'w') as configfile:
+        config.write(configfile)
+    os.chmod(INI_PATH, 0o600)
+
+
 def get_claim_token(dev=False):
 
     setup_endpoints(dev)
@@ -673,12 +703,13 @@ def enroll_device(enroll_token, claim_token, device_id):
             logger.error('Failed to enroll device...')
             _log_request_errors(enroll_req)
             req_error_log('post', 'Enroll by token', enroll_req, caller='enroll-device')
+            return False
         else:
             logger.info('Device {} enrolled successfully.'.format(device_id))
-
+            return True
     except requests.exceptions.RequestException:
         logger.exception("enroll_device :: rises exception:")
-
+        return False
 
 def run(ping=True, dev=False, logger=logger):
 
@@ -692,6 +723,7 @@ def run(ping=True, dev=False, logger=logger):
             write_metadata({'device_id': device_id}, rewrite_file=True)
         else:
             device_id = get_device_id()
+            try_enroll_in_operation_mode(device_id=device_id, dev=dev)
             write_metadata({'device_id': device_id}, rewrite_file=False)
             if not time_for_certificate_renewal() and not is_certificate_expired():
                 if ping:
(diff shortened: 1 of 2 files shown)


eugenosm:
And one question left.
Now it checks enrollment key record only at bootstraping.
May be we need to check it somwhere else? Or there is claim url for other cases?

agent/__init__.py
```diff
@@ -311,6 +311,12 @@ def get_ini_log_file():
     return config['DEFAULT'].get('log_file', None)
 
 
+def get_enroll_token():
+    config = configparser.ConfigParser()
+    config.read(INI_PATH)
+    return config['DEFAULT'].get('enroll_token', None)
```
rptrchv: `.get` returns None by default. so pls stop reproducing this bad design pattern

agent/__init__.py
```diff
@@ -642,6 +648,29 @@ def write_metadata(data, rewrite_file):
     metadata_path.chmod(0o644)
 
 
+def enroll_device(enroll_token, claim_token, device_id):
+
```
rptrchv: pls stop:
1) adding unnecessary blank lines
2) spread quite obvious/simple/short code over multiple lines if it can fit one line

agent/__init__.py
```diff
@@ -642,6 +648,29 @@ def write_metadata(data, rewrite_file):
...
+        'claim_token': claim_token,
+        'device_id': device_id
+    }
+
+    try:
+        enroll_req = requests.post(
+            '{}/v0.2/enroll-device'.format(WOTT_ENDPOINT),
+            json=payload
+        )
+    except requests.exceptions.RequestException:
```
rptrchv: why don't you write the error details to the log?

agent/__init__.py
```diff
@@ -642,6 +648,29 @@ def write_metadata(data, rewrite_file):
...
+        logger.exception("enroll_device :: rises exception:")
+
+    if not enroll_req.ok:
```
rptrchv: `enroll_req` won't be available in case of exception (above)

agent/__init__.py
```diff
@@ -642,6 +648,29 @@ def write_metadata(data, rewrite_file):
...
+        logger.error('Failed to enroll device...')
+        req_error_log('post', 'Enroll by token', enroll_req, caller='enroll-device')
+    else:
+        logger.info(enroll_req.json()['message'])
```
rptrchv: As I said in the task's API side PR, we don't need pass some text over internet to inform client about everything went well

rptrchv:
@eugenosm pls make sure `debian/postinst` displays meaningful messages both in case of manual device claim and automatic enrollment

agent/__init__.py

agent/__init__.py
eugenosm: The first of all (right after good working) code must be lite readable. 
And thus i have a habit to leave blank line after function header, just for regular form. Because some functions have a long args list which is not fits on one line. 
The next one is a dictionaries, with strings, when it wrote line by line this is reading simply than if this distributed on several lines by 3-5 non regular items per line. 
And third one,  yes i have to place space lines between logical or data blocks. I do not understand you. Did you like a glued text? 

Ok then, i could do that way. But this is not good idea in 30-50% of cases.
Fixed.

agent/__init__.py
eugenosm: > exception(msg, *args, **kwargs)¶
Logs a message with level ERROR on this logger. The arguments are interpreted as for debug(). Exception info is added to the logging message. This method should only be called from an exception handler.

using the `logger.exception()` function in this sense is the same as including the `exc_info=True` parameter in the `logger.error()`. 

P.S. I added more suitable error handling function on `not  request.ok`. But here, what else over the exception info, you want to see in log at this place?

agent/__init__.py
eugenosm: Fixed

agent/__init__.py
eugenosm: Fixed

eugenosm:
> eugenosm pls make sure debian/postinst displays meaningful messages both in case of manual device claim and automatic enrollment

I implemented this based on this behavior (i mean as post install process)

This is when device enrollment failed for some reason (it still offers to claim via url/token):
image

And this is when device was enrolled successfully during postinstall process:
image

agent/__init__.py
rptrchv: > using the `logger.exception()` function in this sense is the same as including the `exc_info=True` parameter in the `logger.error()`.

ok, thanks for clarification

agent/__init__.py
rptrchv: > And one question left.
> Now it checks enrollment key record only at bootstraping.
> May be we need to check it somwhere else? Or there is claim url for other cases?

nothing comes to my mind now, but it's your task to check the whole project's codebase for finding an answer

rptrchv:
ok

agent/__init__.py
eugenosm: This is more behavior question than code.  @vpetersson do we need to check for enroll token in config any time else than on bootstrap install?  I assume not, but may be i'm wrong?

vpetersson: That's a good question. I guess we should check it if the device is unclaimed.

eugenosm: Ok. Added check/enroll to "main cycle".

tests/test_agent.py
```diff
@@ -566,6 +566,251 @@ def test_fetch_device_metadata(tmpdir):
         ])
 
 
+def test_enroll_device_ok(tmpdir):
+    executor.Locker.LOCKDIR = str(tmpdir)
+    message = "Device d3d301961e6c4095b59583083bdec290.d.wott-dev.local enrolled successfully."
+
+    mock_resp = mock.Mock()
+    mock_resp.raise_status = 200
```
rptrchv: are you sure this is needed?

eugenosm: not really. fixed.

vpetersson:
@eugenosm Looks like tests are failing on Stretch. Can you investigate?

eugenosm:
About test fail:
https://github.com/WoTTsecurity/agent/issues/195

So this is not relevant to the current task.

PR #189, opened 2019-07-15
Add support for Raspbian Buster

### LINKED ISSUE #185: Add support for Raspbian Buster
The Foundation has moved to Buster now and Jessie is no longer the default installation. We need to make the shift too and make sure the agent both works on Jessie and Buster.

rptrchv:
Closes #185


.circleci/config.yml
```diff
@@ -34,13 +68,21 @@ jobs:
...
       - store_artifacts:
           path: /tmp/deb
+workflows:
+  version: 2
+  build_all:
+    jobs:
+      - build_debian_buster
+      - build_debian_stretch:
+          requires:
+            - build_debian_buster
```
vpetersson: Why do we need to wait here?

.circleci/config.yml
```diff
@@ -1,6 +1,45 @@
...
+            flake8 ./ --exclude=venv --ignore=E501 --output-file test-reports
+      - run:
+          name: Run tests
+          command: |
+            pytest --cov=agent
+      - run:
+          name: Build dpkg packages
+          command: |
+            python3 -c 'import version; version.write_changelog()'
+            debuild -e CIRCLE_BUILD_NUM -i -us -uc -b
```
vpetersson: Shouldn't this be `${CIRCLE_BUILD_NUM}`?

.circleci/config.yml
rptrchv: Just to avoid displaying 'NOT RUNNING' jobs' statuses in the CircleCI UI.

vpetersson: I guess it doesn't matter about parallel mode right now, but it will single thread it later.

rptrchv: > but it will single thread it later

it's not clear from your comment is this ^ a desired or not.

vpetersson: It's fine for now. Let's remove it when we add more.

PR #203, opened 2019-08-20
Build only on Stretch. Push to Buster, Stretch and Xenial.

a-martynovich:
#202


rptrchv:
Why on Stretch, not on Buster which is the only currently supported Raspbian version?

PR #210, opened 2019-08-30
Send a list of installed deb packages.

### LINKED ISSUE #22: Send list of installed packages to API
In order to assess the security state of the installed packages, we need to submit the list upstream. This is of course a rather complicated task, so we need to make some assumptions to limit the scope. In the initial version, let's only worry about Debian/Raspbian.

For instance, we can extract all the installed packages, along with their versions from Raspbian by running:

```
$ sudo dpkg -l
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                                                                                  Version                                         Architecture                                    Description
+++-=====================================================================================-===============================================-===============================================-===============================================================================================================================================================================
ii  adduser                                                                               3.115                                           all                                             add and remove users and groups
ii  alsa-utils                                                                            1.1.3-1                                         armhf                                           Utilities for configuring and using ALSA
ii  apparmor                                                                              2.11.0-3+deb9u2                                 armhf                                           user-space parser utility for AppArmor
ii  apt                                                                                   1.4.9                                           armhf                                           commandline package manager
ii  apt-listchanges                                                                       3.10                                            all                                             package change history notification tool
[...]
```

This is of course not very easily parseable, but serves the purpose of the kind of data we want to send upstream.

a-martynovich:
Closes #22

### DIFF
--- a/agent/__init__.py
+++ b/agent/__init__.py
@@ -29,7 +29,7 @@ from agent import journal_helper
 from agent import rpi_helper
 from agent import security_helper
 from agent.executor import Locker
-from agent.rpi_helper import Confinement, detect_confinement, detect_installation
+from agent.rpi_helper import Confinement, detect_confinement, detect_installation, get_deb_packages
 
 
 CONFINEMENT = detect_confinement()
@@ -367,7 +367,7 @@ def get_uptime():
 
 
 def get_open_ports(dev=False):
-    connections, ports = security_helper.netstat_scan()
+    _, ports = security_helper.netstat_scan()
     return ports
 
 
@@ -388,7 +388,8 @@ def send_ping(dev=False):
         'uptime': get_uptime(),
         'agent_version': str(__version__),
         'confinement': CONFINEMENT.name,
-        'installation': detect_installation().name
+        'installation': detect_installation().name,
+        'deb_packages': get_deb_packages()
     }
 
     # Things we can't do within a Snap or Docker
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
           command: |
             apt update
             apt install -y python3-pip python3-netifaces python3-requests python3-tz python3-iptables python3-systemd \
-            python3-psutil devscripts python3-all ruby-dev
+            python3-psutil python3-apt devscripts python3-all ruby-dev
             pip3 install -r requirements-dev.txt
       - run:
           name: Run linting and metrics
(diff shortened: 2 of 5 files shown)


vpetersson:
@a-martynovich let's add some caching here too (such as the checksum of all installed packages, sorted). That way we don't need to send the entire list of installed packages every time, but rather just do it when it changes.

agent/__init__.py
```diff
@@ -367,7 +367,7 @@ def get_uptime():
 
 
 def get_open_ports(dev=False):
-    connections, ports = security_helper.netstat_scan()
+    _, ports = security_helper.netstat_scan()
     return ports
```
rptrchv: `return security_helper.netstat_scan()[0]`

agent/__init__.py
```diff
@@ -401,7 +405,7 @@ def send_ping(dev=False):
 
     # Things we cannot do in Docker
     if CONFINEMENT not in (Confinement.DOCKER, Confinement.BALENA):
-        blocklist = ping.json()
+        blocklist = ping['firewall']
```
rptrchv: Inconsistent change. See my comment in the api project's related PR

rptrchv:
@a-martynovich 
Does this feature make the project completely inconsistent with all non-apt-based platforms?

vpetersson:
> @a-martynovich
> Does this feature make the project completely inconsistent with all non-apt-based platforms?

Right now, we're only targeting Ubuntu/Debian/Raspbian but in the future, we should probably expand to support other platforms. As such, we should probably have  a conditional check and disable this for non=-debian derived distros.

agent/__init__.py
```diff
@@ -390,6 +392,8 @@ def send_ping(dev=False):
         'confinement': CONFINEMENT.name,
         'installation': detect_installation().name
     }
+    if ping.get('deb_packages_hash') != packages['hash']:
+        payload['deb_packages'] = packages
```
rptrchv: Pls consider my recommendation to always send something in `payload['deb_packages']`

agent/__init__.py
a-martynovich: It's `[1]`. See, you've already made a mistake yourself. I believe we should specify what exactly we want to return.

rptrchv: Yes, I used a wrong index but the solution is quite correct. I don't insist on using it if you prefer to introduce a couple of useless variables for the sake of readability instead of adding a short comment.

PR #214, opened 2019-09-16
Audit config files

### LINKED ISSUE #207: Audit security posture of OpenSSH
OpenSSH is likely one of the most common services to run on an edge device. Getting security right for OpenSSH isn't very hard, but it's easy to miss.

Let's add an audit feature that ensures that the configuration is secure.

Here are the values that we should check and warn if detected:

| Configuration  | Bad Value | Warning Type | Comment | 
| ------------- |:-------------:| -----:| -----:|
| PermitEmptyPasswords | yes | Severe | This is pretty bad.
| PermitRootLogin | yes | Severe | |
| Protocol | 2, 1 or 1 | Warning | There little reason for this to be 1 | 
| PasswordAuthentication | yes | Severe | Users should only use key based authentication |
| AllowAgentForwarding | yes | Warning | This is not ideal as it can be intercepted

We should look for the SSH config in some common locations, such as `/etc/ssh/sshd_config`.

### LINKED ISSUE #213: Report checksum of some config files
*Subtask of wottsecurity/api#400*
```
/etc/password
/etc/shadow
/etc/group
/etc/ssh/sshd_config
```

a-martynovich:
Closes #213 
Closes #207

### DIFF
--- a/agent/security_helper.py
+++ b/agent/security_helper.py
@@ -1,5 +1,7 @@
 import crypt
+from hashlib import sha256
 import socket
+import os
 from pathlib import Path
 from socket import SocketKind
 import spwd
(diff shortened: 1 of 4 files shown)


agent/security_helper.py
```diff
@@ -118,3 +120,74 @@ def selinux_status():
...
+    get their last modified time and SHA256 hash.
+    The same info regarding SSHD_CONFIG_PATH is appended (see audit_sshd below),
+    :return: [{'name': ..., 'sha256': ..., 'last_modified': ...}]
+    """
+
+    def digest_sha256(file_path):
+        h = sha256()
+
+        with open(file_path, 'rb') as file:
+            while True:
```
rptrchv: Why do you read file and calculate hash in chunks instead of doing this with the full file content?

agent/security_helper.py
```diff
@@ -118,3 +120,74 @@ def selinux_status():
...
+                continue
+
+            line_split = line.split(maxsplit=1)
+            if len(line_split) != 2:
+                # skip invalid lines
+                continue
+
+            parameter, value = line_split
+            value = value.strip('"')
+            if parameter in ['PermitEmptyPasswords', 'PermitRootLogin', 'PasswordAuthentication', 'AllowAgentForwarding']\
```
rptrchv: Your solution doesn't take care of default positive values of insecure parameters like `PasswordAuthentication` and `AllowAgentForwarding`.

agent/security_helper.py
```diff
@@ -118,3 +120,74 @@ def selinux_status():
...
+                    and value == 'yes' or\
+               parameter == 'Protocol' and value in ['2,1', '1']:
```
rptrchv: Raspbian keeps `Protocol` in `ssh_config`, not in `sshd_config`

vpetersson: @rptrchv those are two different files. `ssh_config` is for the ssh client, and `sshd_config` is for the ssh server.

rptrchv: right. just found out this in the man page

agent/security_helper.py
a-martynovich: good point, never thought of that

rptrchv: too long line

PR #216, opened 2019-09-18
Audit config files.

a-martynovich:
*in place of accidentally merged #214*

### DIFF
--- a/agent/security_helper.py
+++ b/agent/security_helper.py
@@ -4,6 +4,7 @@ import socket
 import os
 from pathlib import Path
 from socket import SocketKind
+import copy
 
 import psutil
 import spwd
@@ -148,14 +149,17 @@ AUDITED_CONFIG_FILES = [
     '/etc/shadow',
     '/etc/group'
 ]
+
 SSHD_CONFIG_PATH = '/etc/ssh/sshd_config'
-SSHD_CONFIG_DEFAULTS = {
-    'PermitEmptyPasswords': 'no',
-    'PermitRootLogin': 'yes',
-    'PasswordAuthentication': 'yes',
-    'AllowAgentForwarding': 'yes',
-    'Protocol': '2,1'
+# Value: (default, safe).
+SSHD_CONFIG_PARAMS_INFO = {
+    'PermitEmptyPasswords': ['no', 'no'],
+    'PermitRootLogin': ['yes', 'no'],
+    'PasswordAuthentication': ['yes', 'no'],
+    'AllowAgentForwarding': ['yes', 'no'],
+    'Protocol': ['2', '2']
 }
+
 BLOCK_SIZE = 64 * 1024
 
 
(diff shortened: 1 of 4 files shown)


agent/security_helper.py
```diff
@@ -118,3 +120,87 @@ def selinux_status():
...
     return {'enabled': selinux_enabled, 'mode': selinux_mode}
+
+
+AUDITED_CONFIG_FILES = [
+    '/etc/passwd',
+    '/etc/shadow',
+    '/etc/group'
+]
+SSHD_CONFIG_PATH = '/etc/ssh/sshd_config'
+SSHD_CONFIG_DEFAULTS = {
```
rptrchv: Pls check the default values with `man sshd_config`.
And, an interesting observation: looks like sshd doesn't support 'Protocol' config parameter at all now - it was mentioned in the previous versions' config and man page, but completely missing now.

a-martynovich: Checked, doesn't match the [old docs](https://linux.die.net/man/5/sshd_config). For instance the default for `PermitRootLogin` is now `prohibit-password` and as you said `Protocol` is removed.
@vpetersson what do we do with that?

vpetersson: Well, i take it we need to support it still, but perhaps make it conditional based on the version? Unfortunately this makes it a bit more messy. What version did this change in? I would imagine OpenSSH to be very backward compatible.

a-martynovich: These defaults are the same in Stretch and Buster. The [internet man](https://linux.die.net/man/5/sshd_config) is dated 2013 so it may be relevant to Jessie. I'm now checking mans in Jessie.

a-martynovich: In Jessie the default value for Protocol is "2" and PermitRootLogin is "yes".
So... I guess we need conditional handing here.

agent/security_helper.py
```diff
@@ -148,14 +149,17 @@ def selinux_status():
...
-    'PermitEmptyPasswords': 'no',
-    'PermitRootLogin': 'yes',
-    'PasswordAuthentication': 'yes',
-    'AllowAgentForwarding': 'yes',
-    'Protocol': '2,1'
+# Value: (default, safe).
+SSHD_CONFIG_PARAMS_INFO = {
+    'PermitEmptyPasswords': ['no', 'no'],
+    'PermitRootLogin': ['yes', 'no'],
+    'PasswordAuthentication': ['yes', 'no'],
```
a-martynovich: In some newer versions the default for PasswordAuthentication is "no". See for example
https://www.freebsd.org/cgi/man.cgi?query=sshd_config&apropos=0&sektion=5&manpath=FreeBSD+12.0-RELEASE&arch=default&format=html
Also in older versions the default for Protocol is "2,1".

a-martynovich:
LGTM

PR #225, opened 2019-09-26
Show Process information in dashboard for ports

rptrchv:
Agent-side companion task for https://github.com/WoTTsecurity/api/issues/415


agent/security_helper.py
```diff
@@ -62,6 +74,7 @@ def netstat_scan():
             'port': c.laddr[1],
             'proto': {SocketKind.SOCK_STREAM: 'tcp', SocketKind.SOCK_DGRAM: 'udp'}.get(c.type),
             'state': c.status if c.type == socket.SOCK_STREAM else None,
+            'process_info': get_process_info(c.pid)
```
a-martynovich: we already send the full list of processes to the server, why this extra field?

rptrchv: > we already send the full list of processes to the server, why this extra field?

fixed

PR #227, opened 2019-09-27
AppArmor, SELinux, Pi detection, Bionic support

### LINKED ISSUE #211: Add support for Ubuntu 18.04 (Bionic)
We should expand our x86 support to include Ubuntu 18.04 as well, as it is the latest LTS release.

### LINKED ISSUE #224: Check AppArmor and SELinux status in Snap
Currently we check this by running AppArmor and SELinux executables which is forbidden in Snap. Other methods are available, for example we can use python3-apparmor and python3-selinux packages (only available as deb).

Try using these packages inside Snap and see if they are allowed to do their job (at least read-only). If not, devise another way of getting AppArmor/SELinux status.

### LINKED ISSUE #226: Improve Pi detection
`/proc/cpuinfo` in Ubuntu Core and Server does not contain "Hardware" and "Serial" fields. Instead some other files should be queried.

To get device model we can read `/proc/device-tree/model`. It contains data like `Raspberry Pi 3 Model B Plus Rev 1.3\0` (notice the null byte at the end). Serial number is at `/proc/device-tree/serial-number`. Also `/proc/device-tree/compatible` contains info in such format: `raspberrypi,3-model-b-plus brcm,bcm2837`.

a-martynovich:
Closes #226 
Closes #224 
Closes #211 
**Requires** wottsecurity/api#426

### DIFF
--- a/agent/security_helper.py
+++ b/agent/security_helper.py
@@ -84,15 +84,25 @@ def is_app_armor_enabled():
     """
     try:
         import LibAppArmor
-        return LibAppArmor.aa_is_enabled() == 1
     except ImportError:
+        # If Python bindings for AppArmor are not installed (if we're
+        # running on Jessie where we can't build python3-apparmor package)
+        # we resort to calling aa-status executable.
         try:
             from sh import aa_status
         except ImportError:
             return False
 
-        # Returns 0 if enabled and 1 if disabled
-        return aa_status(['--enabled'], _ok_code=[0, 1]).exit_code != 1
+        # Return codes (as per aa-status(8)):
+        # 0   if apparmor is enabled and policy is loaded.
+        # 1   if apparmor is not enabled/loaded.
+        # 2   if apparmor is enabled but no policy is loaded.
+        # 3   if the apparmor control files aren't available under /sys/kernel/security/.
+        # 4   if the user running the script doesn't have enough privileges to read the apparmor
+        #    control files.
+        return aa_status(['--enabled'], _ok_code=[0, 1, 2, 3, 4]).exit_code in [0, 2]
+    else:
+        return LibAppArmor.aa_is_enabled() == 1
 
 
 def selinux_status():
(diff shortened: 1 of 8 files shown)


agent/security_helper.py
```diff
@@ -83,38 +83,45 @@ def is_app_armor_enabled():
     Returns a True/False if AppArmor is enabled.
     """
     try:
-        from sh import aa_status
+        import LibAppArmor
+        return LibAppArmor.aa_is_enabled() == 1
```
rptrchv: pls move this out of `try`

agent/security_helper.py
```diff
@@ -83,38 +83,45 @@ def is_app_armor_enabled():
...
     except ImportError:
-        return False
+        try:
+            from sh import aa_status
+        except ImportError:
```
rptrchv: pls comment what all execution branches mean

agent/security_helper.py
```diff
@@ -83,38 +83,45 @@ def is_app_armor_enabled():
...
+            return False
 
-    # Returns 0 if enabled and 1 if disable
-    get_aa_status = aa_status(['--enabled'], _ok_code=[0, 1]).exit_code
-    if get_aa_status == 1:
-        return False
-    return True
+        # Returns 0 if enabled and 1 if disabled
+        return aa_status(['--enabled'], _ok_code=[0, 1]).exit_code != 1
```
rptrchv: it's not clear why didn't you put **all** possible return codes of `aa_status` to `_ok_code`

agent/security_helper.py
```diff
@@ -83,38 +83,45 @@ def is_app_armor_enabled():
...
         {'enabled': False, 'mode': 'enforcing'}
     """
-    selinux_enabled = None
+    selinux_enabled = False
     selinux_mode = None
 
     try:
-        from sh import sestatus
+        import selinux
+        if selinux.is_selinux_enabled() == 1:
```
rptrchv: pls move this branch out of `try`

agent/security_helper.py
a-martynovich: fair enough, needs fixing

agent/security_helper.py
rptrchv: nope. e.g. it's not clear what does it mean if `LibAppArmor` isn't available for import

a-martynovich: ok I'll add a comment

rptrchv:
@a-martynovich and pls fix the Codacy warning

a-martynovich:
@rptrchv Does it make sense to you? To me it doesn't. If I remove `self` from there it will be impossible to call this method on an instance (`is_file() takes 0 positional arguments but 1 was given`).
image

rptrchv:
> @rptrchv Does it make sense to you? To me it doesn't.

I've already put this exact pylint check to the skip-list in my recent task. Don't do anything re this. I wrote my comment here without looking into the error

PR #228, opened 2019-10-02
Integrate with agent-snap

a-martynovich:


vpetersson:
@a-martynovich can we do an exclude on this such that it isn't included in the debian package build? (or perhaps it isn't already).

a-martynovich:
@vpetersson debian packaging process runs `python setup.py install` and packs the result of that. Therefore nothing that wasn't installed by `setup.py` goes into deb.

PR #239, opened 2019-10-16
Check for auto upgrades

a-martynovich:
wottsecurity/api#445


rptrchv:
@a-martynovich what does this Codacy/PR Quality Review check fail mean? why it didn't manage to import `apt_pkg`?

a-martynovich:
@rptrchv because it's stupid and doesn't know that apt_pkg is part of python-apt.

rptrchv:
> @rptrchv because it's stupid and doesn't know that apt_pkg is part of python-apt.
ok, I'm approving the PR if you think we can ignore this

PR #243, opened 2019-10-22
Check for mysql root access.

a-martynovich:
wottsecurity/api#456

### DIFF
--- a/agent/security_helper.py
+++ b/agent/security_helper.py
@@ -1,10 +1,11 @@
+import copy
 import crypt
-from hashlib import sha256
-import socket
 import os
+import socket
+import subprocess
+from hashlib import sha256
 from pathlib import Path
 from socket import SocketKind
-import copy
 
 import psutil
 import spwd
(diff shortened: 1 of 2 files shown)


agent/security_helper.py
```diff
@@ -249,3 +250,17 @@ def audit_sshd():
         if insecure_params[param] != config[param][1]:
             issues[param] = insecure_params[param]
     return issues
+
+
+def mysql_root_access():
+    try:
+        subprocess.check_call(["mysql", "-uroot", "-eSHOW DATABASES;"], timeout=5,
+                              stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        return True
```
rptrchv: pls move this out of the `try` clause

a-martynovich: Why? What's going to change?

rptrchv: In this particular case - nothing besides readability. But in general a good practice is to not put in `try` anything besides the code supposed to throw an exception in some condition

a-martynovich: Splitting these two lines into two blocks, one of which will be at the bottom, will decrease readability. For no reason whatsoever.

rptrchv: I don't agree, but let's forget about subjective readability. What about following good practice rules?

PR #245, opened 2019-10-31
Detect cpu vulnerabilities and vendor.

a-martynovich:
wottsecurity/api#482


rptrchv:
`platform.linux_distribution()` doesn't return a codename on our target distros. isn't it a problem?

a-martynovich:
@rptrchv It is used as a last resort if `/etc/os-release` is unavailable. On Ubuntu it returns `('Ubuntu', '18.04', 'bionic')`. If it doesn't give us codename then we're probably on an unsupported system.

agent/rpi_helper.py
```diff
@@ -144,3 +145,43 @@ def auto_upgrades_enabled():
...
+    """
+    Finds which currently installed deb package contains the currently running kernel.
+    :return:
+        If the currently running kernel was installed by deb package:
+            a dict where 'source_name' and 'source_version' are the same as returned by get_deb_packages().
+        Otherwise: None
+    """
+    import apt
+
+    boot_image = kernel_cmdline().get('BOOT_IMAGE')
```
rptrchv: My RPI3 default installation doesn't have the 'BOOT_IMAGE' kernel command line parameter. what does that mean?

a-martynovich: It means that it boots differently, not with a vmlinuz image, I wrote about it [here](https://github.com/WoTTsecurity/api/issues/482#issuecomment-546829943). Pi is not vulnerable anyway, since it's ARM.

rptrchv:
> If it doesn't give us codename then we're probably on an unsupported system.

Debian/Raspbian of any version are unsupported systems?

a-martynovich:
On Debian/Raspbian /etc/os-release is present.

rptrchv:
> On Debian/Raspbian /etc/os-release is present.
> And why are you asking this question? The PR with these changes was merged last week.

Because this functionality is crucial for this task's server side solution

a-martynovich:
@rptrchv again, what do you think is wrong?

agent/rpi_helper.py
```diff
@@ -144,3 +145,43 @@ def auto_upgrades_enabled():
...
         return False
+
+
+def kernel_cmdline():
+    """
+    Parses kernel parameters (aka cmdline).
+    :return: A dict where 'name' is kernel parameter name and 'value' is its value or empty string if no value provided.
+    """
+    cmdline_path = Path('/proc/cmdline')
+    cmdline_matches = re.compile(r"([\w\-\.]+)(\=(\"[\w\W]+\"|[\w\S]+)?)?").findall(cmdline_path.read_text())
```
rptrchv: I recommend to move the re pattern compilation out of the function and use already compiled instance in it.

agent/rpi_helper.py
```diff
@@ -144,3 +145,43 @@ def auto_upgrades_enabled():
...
+
+def kernel_deb_package():
+    """
+    Finds which currently installed deb package contains the currently running kernel.
+    :return:
+        If the currently running kernel was installed by deb package:
+            a dict where 'source_name' and 'source_version' are the same as returned by get_deb_packages().
+        Otherwise: None
+    """
+    import apt
```
rptrchv: why do you import it here?

a-martynovich: Because it's not available in snap

agent/rpi_helper.py
a-martynovich: It is going to only be used twice during the whole lifetime of the program. And it's going to parse a string which is around 100 chars long. No need to optimize anything.

agent/rpi_helper.py
```diff
@@ -144,3 +145,43 @@ def auto_upgrades_enabled():
...
+
+    boot_image = kernel_cmdline().get('BOOT_IMAGE')
+    if not boot_image:
+        return
+
+    class FileFilter(apt.cache.Filter):
```
rptrchv: what's the purpose of storing this class definition inside the function?

a-martynovich: because it's only used inside this function

agent/security_helper.py
```diff
@@ -264,3 +267,67 @@ def mysql_root_access():
...
+        return res
+
+    sys_vulnerabilities = Path('/sys/devices/system/cpu/vulnerabilities')
+    if not sys_vulnerabilities.is_dir():
+        # Directory does not exist: either smth is bind-mounted over it or the kernel is too old.
+        vulnerable = None
+    else:
+        vulnerable = False
+        for name in ('l1tf', 'mds', 'meltdown', 'spectre_v1', 'spectre_v2', 'spec_store_bypass'):
+            f = sys_vulnerabilities / name
```
rptrchv: pls choose more meaningful variables names

agent/security_helper.py
```diff
@@ -264,3 +267,67 @@ def mysql_root_access():
...
+            if f.exists():
```
rptrchv: I'd recommend to check for `.is_file()`

agent/security_helper.py
```diff
@@ -264,3 +267,67 @@ def mysql_root_access():
...
+                # If CPU is not prone to this vulnerability the status file will start with
+                # 'Not affected' or 'Mitigation: ...'. Otherwise it will start with 'Vulnerable: ...'.
+                if f.read_text().startswith('Vulnerable'):
+                    vulnerable = True
+                    break
+            else:
+                # Status file does not exist: smth is bind-mounted over it or the kernel is not completely patched.
+                vulnerable = None
+                break
```
rptrchv: Why break? Aren't you going to check other vulns' files?

a-martynovich: No, one is enough already.

rptrchv:
> @rptrchv again, what do you think is wrong?

Nothing. I just needed to clarify things for myself

agent/rpi_helper.py
rptrchv: As you wish. But fixing this would take you less time than writhing the reply ^.
The meaning of re pattern compilation is to do it once and use multiple times. If you believe it's useless - them you can do `findall` without compilation at all

agent/rpi_helper.py
rptrchv: it's suboptimal and not pythonic, but ok, I don't have too much spare time discussing this

PR #246, opened 2019-11-05
Self-update systemd timer.

### LINKED ISSUE #244: Self-update and/or ability to add new checks to existing agents
Note: systemd timer for upgrade only wott-agent

a-martynovich:
Closes #244

### DIFF
--- a/debian/wott-agent-self-update.service
+++ b/debian/wott-agent-self-update.service
@@ -3,5 +3,5 @@ Description=wott-agent self update
 
 [Service]
 Type=oneshot
-ExecStart=/opt/wott/wott-agent-self-update.sh
+ExecStart=/opt/wott/bin/wott-agent-self-update.sh
 RemainAfterExit=True
--- a/debian/wott-agent-self-update.service
+++ b/debian/wott-agent-self-update.service
@@ -3,5 +3,5 @@ Description=wott-agent self update
 
 [Service]
 Type=oneshot
-ExecStart=/usr/local/bin/wott-agent-self-update.sh
+ExecStart=/opt/wott/wott-agent-self-update.sh
 RemainAfterExit=True
--- a/debian/wott-agent-self-update.service
+++ b/debian/wott-agent-self-update.service
@@ -1,0 +1,7 @@
+[Unit]
+Description=wott-agent self update
+
+[Service]
+Type=oneshot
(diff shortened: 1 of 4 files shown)


debian/wott-agent-self-update.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
```
vpetersson: Add in 

```
# vim: tabstop=4 shiftwidth=4 softtabstop=4
# -*- sh-basic-offset: 4 -*-

set -euo pipefail
IFS=$'\n\t'
```

debian/wott-agent-self-update.sh
```diff
@@ -0,0 +1,4 @@
...
+apt-get update
+apt-get install wott-agent
```
vpetersson: This won't work. It needs the `-y` flag, otherwise it will just throw a dialogue. Change it to `apt-get install -ys wott-agent`

debian/wott-agent-self-update.service
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=wott-agent self update
+
+[Service]
+Type=oneshot
+ExecStart=/opt/wott/wott-agent-self-update.sh
```
vpetersson: Let's create a `bin` directory under `/opt/wott` as we will likely need more scripts in the future.

debian/wott-agent-self-update.sh
a-martynovich: Why do we need to set `IFS` here and why `\n\t`?

vpetersson: It's just something that I've used over the years. I forgot the exact reason for this, but in general it should harden the bash scripts.

a-martynovich: This is not only useless in our case but also harmful. Let's not do this.

vpetersson: Care to elaborate?

a-martynovich: It affects tokenizing. You don't expect different tokenizing when you're writing a script unless you really need this.
Default value for IFS is `' \t\n'`. Why would you need to change it?

vpetersson: Alright, fair, let's leave IFS as is.

PR #253, opened 2019-12-19
Adds new installation method 

vpetersson:
To be used either by `curl` or with directly from the repo.


install.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+mkdir -p /opt/wott
+
+if [ ! -f /etc/wott/config.ini ]; then
+    echo -e "[DEFAULT]\\nenroll_token = $CLAIM_TOKEN" > /opt/wott/config.ini
+fi
+
+# If Ubuntu/Debian is detected
+if [ -f /etc/lsb-release ]; then
```
a-martynovich: Will not work on Jessie and probably some other flavors. It's better to parse `/etc/os-release` instead, like we do this in Python code.

install.sh
```diff
@@ -0,0 +1,28 @@
...
+
+if [[ $(id -u) -ne 0 ]]; then
+    echo "Please run as root or using sudo"
+    exit 1
+fi
+
+mkdir -p /opt/wott
+
+if [ ! -f /etc/wott/config.ini ]; then
+    echo -e "[DEFAULT]\\nenroll_token = $CLAIM_TOKEN" > /opt/wott/config.ini
```
a-martynovich: You're checking for `/etc/wott/config.ini` and then writing to `/opt/wott/config.ini`. Also where is `CLAIM_TOKEN` supposed to come from?

install.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# vim: tabstop=4 shiftwidth=4 softtabstop=4
+# -*- sh-basic-offset: 4 -*-
```
a-martynovich: Is there any practical reason for setting *your* preferred Vim and Emacs settings in every bash script? Please don't tell me it's a "good practice".

install.sh
vpetersson: Do we need to check both, or is `/etc/os-release` sufficient for all?

install.sh
vpetersson: I'm checking to make sure it doesn't exist - and only if it *doesn't* exist, i write to it.

Claim token comes from an environment variable as mentioned in Slack.

a-martynovich: Please read again. You're checking for `/etc/wott/config.ini` but writing to `/opt/wott/config.ini`.

install.sh
a-martynovich: `/etc/os-release` is sufficient for all.

install.sh
vpetersson: This is why we do peer reviews :)

bin/install.sh
```diff
@@ -0,0 +1,34 @@
...
+    fi
+
+    mkdir -p /opt/wott
+
+    if [ ! -f /opt/wott/config.ini ]; then
+        echo -e "[DEFAULT]\\nenroll_token = $CLAIM_TOKEN" > /opt/wott/config.ini
+    fi
+
+    # If Ubuntu/Debian is detected
+    if [ -f /etc/os-release ]; then
```
a-martynovich: Not *test* it, *parse* it

vpetersson: Ok i've refactored it a fair bit now.

vpetersson:
@a-martynovich this looks pretty solid now. Tested it across a number of distributions:

```
> $ for i in ubuntu:14.04 ubuntu:16.04 ubuntu:18.04 debian:8 debian:9 debian:10 centos:6 centos:7 centos:8 fedora; do docker run -v $(pwd):/app -e CLAIM_TOKEN=foobar --rm -ti $i /app/install.sh; done
Detected ubuntu
Detected ubuntu
Detected ubuntu
Detected debian
Detected debian
Detected debian
Unable to detect distribution. Exiting.
Detected CentOS. This is currently an unsupported distribution.
Detected CentOS. This is currently an unsupported distribution.
Detected Fedora. This is currently an unsupported distribution.
```

Issue #257, opened 2020-01-10
Implement functinality to upgrade vulnerable packages


vpetersson:
We want to make it easy for users to patch their systems by having a simple one-liner to do this. Unfortunately, this can become somewhat troublesome when you have different configurations across the fleet. 

As such, we want to extend the CLI for easier patching of packages. 

We want to keep this as dynamic as possible. Hence, we could do something similar to `wott-agent patch $ID`, where $ID is the issue from the backend. This could either be CVE or a change.

vpetersson:
We want to demo this for an investor, so let's take some shortcuts to get the alpha version of this working. Here are some shortcuts:

 * Instead of making an API call back to the back-end to lookup what an ID means, let's use names (e.g. `wott-agent patch openssh-disable-password-auth`).
 * Let's limit the scope to the ssh recommended actions.
 * Let's make sure to run a ping at the end of the run.
 * Let's make sure to take a backup of the file prior to the change.

vpetersson:
MVP is done Sprint 16. We've received great feedback on this. Now we need to revisit the original scope in Sprint 17.

vpetersson:
We need to resume work on this. As next steps, I want to do the following:

- [ ] Port all CVE instructions to the agent (https://github.com/WoTTsecurity/api/issues/771). This needs to work for both Ubuntu and AWS Linux. 
- [ ] Hide 'Certificate expires...' when using `patch`.
- [ ] Display the actual `diff` as per the instructions in the initial ticket.
- [ ] Implement patch for Spectr/Heartbleed RA
- [ ] Implement patch for 'Automatic security updates not enabled'

rptrchv:
13 complexity points

a-martynovich:
20

PR #263, opened 2020-02-26
Patch command (OpenSSH issues only)

a-martynovich:
*Note*: all "actions" are now argparse commands which can have subcommands:
```
$ agent --help
usage: wott-agent [-h] [--dev] [--debug]
                  {whoami,portscan,test-cert,claim-token,claim-url,daemon,node-metadata,credentials,patch}
                  ...

Let's Encrypt for IoT (with more bells and whistles).
When called without arguments, generates node certificate
or renews it if necessary.

positional arguments:
  {whoami,portscan,test-cert,claim-token,claim-url,daemon,node-metadata,credentials,patch}
                        Choose one of the following:
    whoami              Print node ID.
    portscan            Print open ports.
    test-cert           Validate node certificate.
    claim-token         Print claim token.
    claim-url           Print claim URL.
    daemon              Run as daemon
    node-metadata       Fetch node specific, secret metadata.
    credentials         Fetch credentials.
    patch               patch the system

optional arguments:
  -h, --help            show this help message and exit
  --dev                 Developer mode: work with locally running server.
  --debug               Debug mode: set log level to DEBUG.
```

Only "patch" command has a subcommand:
```
$ agent patch --help
usage: wott-agent patch [-h] patch_name

positional arguments:
  patch_name  One of the following:
              openssh-empty-password    OpenSSH: Disable logins with empty password
              openssh-root-login        OpenSSH: Disable root login
              openssh-password-auth     OpenSSH: Disable password authentication
              openssh-agent-forwarding  OpenSSH: Disable agent forwarding
              openssh-protocol          OpenSSH: Force protocol version 2

optional arguments:
  -h, --help  show this help message and exit
```

In order to pass a  dev or debug flag the `--dev` or `--debug` argument should come **before** the command, for example:
```
$ agent --dev whoami
...
$ agent --debug --dev patch openssh-root-login
```

#257


vpetersson:
I guess we also need a separate PR for the updated RAs.

agent/security_helper.py
```diff
@@ -336,3 +340,32 @@ def cpu_vulnerabilities():
         res['mitigations_disabled'] = mitigations_disabled
 
     return res
+
+
+def patch_sshd_config(patch_param, safe_value):
```
vpetersson: Can we backup the file first?

a-martynovich: It gets backed up before and if it gets changed. Look at the end of the function.

agent/__main__.py
```diff
@@ -57,6 +72,9 @@ def main():
     elif args.action == 'daemon':
         logger.info("start in daemon mode...")
         run_daemon(dev=args.dev)
+    elif args.action == 'patch':
+        patch(args.patch_name, dev=args.dev)
+        run(ping=True, dev=args.dev)
```
vpetersson: It should in theory be safe to run `service ssh restart` after, as it has a check-config first:

```
  restart)
	check_for_upstart 1
	check_privsep_dir
	check_config
	log_daemon_msg "Restarting OpenBSD Secure Shell server" "sshd" || true
	start-stop-daemon --stop --quiet --oknodo --retry 30 --pidfile /var/run/sshd.pid
	check_for_no_start log_end_msg
	check_dev_null log_end_msg
	if start-stop-daemon --start --quiet --oknodo --pidfile /var/run/sshd.pid --exec /usr/sbin/sshd -- $SSHD_OPTS; then
	    log_end_msg 0 || true
	else
	    log_end_msg 1 || true
	fi
	;;
```

a-martynovich: Sorry forgot about that

agent/security_helper.py
```diff
@@ -336,3 +340,32 @@ def cpu_vulnerabilities():
...
+            if len(line_split) != 2:
+                continue
+            param, value = line_split
+            value = value.strip('"')
+            if param == patch_param and value != safe_value:
+                logger.info('{}: replacing "{}" with "{}"'.format(param, value, safe_value))
+                patched_lines[-1] = param + ' ' + safe_value + '\n'
+                replaced = True
+        if replaced:
+            shutil.copy(SSHD_CONFIG_PATH, SSHD_CONFIG_PATH + '.bup')
```
vpetersson: Can we copy this file to `/opt/wott/backups/sshd_config.$TIMESTAMP` or similar instead?

agent/security_helper.py
```diff
@@ -336,3 +340,32 @@ def cpu_vulnerabilities():
...
+                # skip empty lines and comments
+                continue
+            line_split = line.split(maxsplit=1)
+            if len(line_split) != 2:
+                continue
+            param, value = line_split
+            value = value.strip('"')
+            if param == patch_param and value != safe_value:
+                logger.info('{}: replacing "{}" with "{}"'.format(param, value, safe_value))
+                patched_lines[-1] = param + ' ' + safe_value + '\n'
```
vpetersson: Can we annotate the end of this line with something like `# Added by wott-agent on $TIMESTAMP`?

agent/__main__.py
vpetersson: Alternatively, we could check it manually (which is what `check_config` does above):

`/usr/sbin/sshd $SSHD_OPTS -t`

And then send a SIGHUP to it. 

That way it becomes distro agnostic.

a-martynovich: SIGHUP won’t restart it, it will reload. This is what “service sshd reload” does. Also I think (need to double-check) reload will fail if config is invalid.

agent/security_helper.py
a-martynovich: We can only add a comment above or below the line, not to the line itself.

vpetersson: This is not true:

```
$ grep Proto /etc/ssh/sshd_config
Protocol 2 #This is a comment
$ sudo service ssh restart
$ sudo /usr/sbin/sshd -t
$ echo $?
0
```

vpetersson: That said, I agree that praxis is to add it above, so maybe we should stick to that.

agent/__main__.py
vpetersson: Yes, `service ssh reload` will do the same, however, the reason for sending SIGHUP vs using `service` is that we don't need to worry about distro specific things (like if it's called "ssh" or "sshd")

a-martynovich: Are we sure that it is in /usr/sbin in all distros?

vpetersson: Nope. That said, I doesn't seem like the `-t` argument requires an absolute path:
```
$ sudo sshd -t
$ echo $?
0
```

(Launching sshd will however require an absolute path.)

vpetersson:
Doesn't seem like PasswordAuthentication patch works when the line is commented out:

```
$ sudo wott-agent patch openssh-password-auth
[...]
patch "PasswordAuthentication"
<Command '/usr/sbin/sshd --hel...(1 more)' call_args {'err_bufsize': 1, '...(495 more)>: starting process
<Command '/usr/bin/lscpu' call_args {'err_bufsize': 1, '...(495 more)>: starting process
Certificate expires in 3 days and 6 hours. No need for renewal.Renewal threshold is set to 3 days.

$ grep PasswordAuth /etc/ssh/sshd_config
#PasswordAuthentication yes
# PasswordAuthentication.  Depending on your PAM configuration,
# PAM authentication, then enable this but set PasswordAuthentication

$ sudo sshd -T  | grep passwordauth
passwordauthentication yes
```

The same is true if a value isn't set:

```
$ grep AllowAgentForwarding /etc/ssh/sshd_config

$ sudo wott-agent patch openssh-agent-forwarding
[...]
patch "AllowAgentForwarding"
<Command '/usr/sbin/sshd --hel...(1 more)' call_args {'decode_errors': 's...(495 more)>: starting process
<Command '/usr/bin/lscpu' call_args {'decode_errors': 's...(495 more)>: starting process
Certificate expires in 3 days and 6 hours. No need for renewal.Renewal threshold is set to 3 days.

$ sudo sshd -T  | grep allowagentforwarding
allowagentforwarding yes
```

PR #269, opened 2020-03-04
Amazon linux

### LINKED ISSUE #250: Add support for Amazon Linux
We've had requests for adding support for [Amazon Linux](https://aws.amazon.com/amazon-linux-2/) in the agent. While Amazon is pitching it as a separate distro, it's CentOS/Fedora derived from the start, meaning that it's using RPM packages.

The starting point would be to build out CentOS support and then make sure these package also work on Amazon Linux.

More data:
* [What Linux distribution is the Amazon Linux AMI based on?](https://serverfault.com/questions/798427/what-linux-distribution-is-the-amazon-linux-ami-based-on)

a-martynovich:
closes #250


a-martynovich:
@rptrchv please fix merge conflicts

agent/os_helper.py
```diff
@@ -0,0 +1,269 @@
...
+    if isinstance(agent.__version__, pkg_resources.Distribution):
+        return Installation.PYTHON_PACKAGE
+    return Installation.NONE
+
+
+def get_packages():
+    # For apt-based distros.
+    try:
+        import apt
+    except ModuleNotFoundError:
```
a-martynovich: Instead of assuming that we're on an rpm-based system if we fail to import `apt` we should simply check if the system is a Debian or not (that's what we have `os_release` for). Otherwise, if for some reason `python-apt` is not installed on a Debian system this code will then try to import `rpm` which doesn't make any sense.

agent/os_helper.py
```diff
@@ -0,0 +1,269 @@
...
+
+def auto_upgrades_enabled():
+    """
+    Checks if auto-updates are enabled on a system.
+    :return: boolean
+    """
+    # For apt-based distros.
+    try:
+        import apt_pkg
+    except ModuleNotFoundError:
```
a-martynovich: same as above

agent/os_helper.py
```diff
@@ -0,0 +1,269 @@
...
+            a dict where 'source_name' and 'source_version' are the same as returned by get_packages().
+        Otherwise: None
+    """
+    boot_image = kernel_cmdline().get('BOOT_IMAGE')
+    if not boot_image:
+        return
+    # For apt-based distros.
+    try:
+        import apt
+    except ModuleNotFoundError:
```
a-martynovich: same as above

agent/os_helper.py
```diff
@@ -0,0 +1,269 @@
...
+    return Installation.NONE
+
+
+def get_packages():
+    # For apt-based distros.
+    try:
+        import apt
+    except ModuleNotFoundError:
+        # For rpm-based distros.
+        import rpm
```
vpetersson: Are we sure this package is always install on RHEL/CentOS/AWS Linux?

a-martynovich: Yes, we have a requirement for it

agent/os_helper.py
```diff
@@ -0,0 +1,269 @@
...
+    Returns a dict with the following items:
+    distro: Concrete distro name. Examples: raspbian, ubuntu, debian, ubuntu-core.
+    version: Short, numerical version. Examples: 9, 18.04, 18.
+    distro_root: The root distro (from which the distro was branched). Optional. Examples: debian.
+    full_version: Longer, human-readable version. Optional. Examples (last one is from ubuntu-core):
+        "9 (stretch)", "18.04.3 LTS (Bionic Beaver)", 18
+    codename: Distro version codename. Optional. Examples: stretch, bionic.
+    """
+
+    os_release = Path('/etc/os-release')
```
vpetersson: Example of `/etc/os-release` on AWS Linux:


```
$ cat /etc/os-release
NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"
```

a-martynovich: and?

vpetersson: And nothing. Just as a reference.

PR #270, opened 2020-03-04
Check if the auditd package is installed

rptrchv:
Connected to https://github.com/WoTTsecurity/api/issues/658


a-martynovich:
We can do this on backend

PR #273, opened 2020-03-05
Amazon: Change binary install path, run wott-agent after install.

a-martynovich:


bin/wott-agent.sh
```diff
@@ -0,0 +1 @@
+/usr/bin/wott-agent
```
vpetersson: Shouldn't this be `/usr/local/bin/wott-agent`?

a-martynovich: No, it’s correct

PR #276, opened 2020-03-09
Detect and send kernel metapackages

### LINKED ISSUE #268: Detect and send kernel metapacakages
Add new fields to ping: `kernel_metapackage` and `kernel_headers_metapackage`. Detect them as follows.

Let's say the currently running kernel is in `linux-image-4.15.0-1060-aws`. Another package, `linux-image-aws`, depends on it. Another package, `linux-aws`, depends on `linux-image-aws`. Dependency chain: `linux-aws => linux-image-aws => linux-image-4.15.0-1060-aws`. I believe nothing depends on `linux-aws` (need to double-check). `linux-aws` is the kernel meta-package.

`linux-headers-4.15.0-1060-aws` is listed as a recommendation for  `linux-image-4.15.0-1060-aws`. Dependency chain: `linux-headers-aws => linux-headers-4.15.0-1060-aws => linux-aws-headers-4.15.0-1060`. `linux-headers-aws` is a meta-package.

rptrchv:
Closes #268

### DIFF
--- a/agent/os_helper.py
+++ b/agent/os_helper.py
@@ -7,7 +7,7 @@ from pathlib import Path
 from enum import Enum
 import pkg_resources
 
-DEBIAN_KERNEL_PKG_NAME_RE = re.compile(r'(.+-\d+.\d+.\d+-)(\d+)([\.-].+)')
+DEBIAN_KERNEL_PKG_NAME_RE = re.compile(r'(.+-\d+\.\d+\.\d+-)(\d+)([.-].+)')
 
 
 def detect_raspberry_pi():
--- a/agent/os_helper.py
+++ b/agent/os_helper.py
@@ -7,7 +7,7 @@ from pathlib import Path
 from enum import Enum
 import pkg_resources
 
-import agent
+debian_kernel_pkg_name_re = re.compile(r'(.+-\d+.\d+.\d+-)(\d+)(-.+)')
 
 
 def detect_raspberry_pi():
@@ -94,6 +94,7 @@ def detect_installation():
                 if __file__.encode() in package_header[rpm.RPMTAG_FILENAMES]:
                     return Installation.RPM
     # Other.
+    import agent
     if isinstance(agent.__version__, pkg_resources.Distribution):
         return Installation.PYTHON_PACKAGE
     return Installation.NONE
@@ -253,6 +254,20 @@ def kernel_cmdline():
     return {name: value.strip('"') for name, _, value in cmdline_matches}
 
 
+def get_kernel_deb_package(boot_image_path):
+    import apt
+
+    class FileFilter(apt.cache.Filter):
+        def apply(self, pkg):
+            return pkg.is_installed and boot_image_path in pkg.installed_files
+
+    cache = apt.cache.FilteredCache(apt.Cache())
+    cache.set_filter(FileFilter())
+    kernel_debs = list(cache)
+    if kernel_debs:
+        return kernel_debs[0]
+
+
 def kernel_package():
     """
     Finds which currently installed deb package contains the currently running kernel.
@@ -261,35 +276,26 @@ def kernel_package():
             a dict where 'source_name' and 'source_version' are the same as returned by get_packages().
         Otherwise: None
     """
-    boot_image = kernel_cmdline().get('BOOT_IMAGE')
-    if not boot_image:
+    boot_image_path = kernel_cmdline().get('BOOT_IMAGE')
+    if boot_image_path is None:
         return
-    # For apt-based distros.
     if is_debian():
-        import apt
-
-        class FileFilter(apt.cache.Filter):
-            def apply(self, pkg):
-                return pkg.is_installed and boot_image in pkg.installed_files
-
-        cache = apt.cache.FilteredCache(apt.Cache())
-        cache.set_filter(FileFilter())
-        kernel_deb = list(cache)
-        if kernel_deb:
-            kernel_pkg = kernel_deb[0].installed
+        # For apt-based distros.
+        kernel_pkg = get_kernel_deb_package(boot_image_path)
+        if kernel_pkg is not None:
             return {
-                'name': kernel_pkg.package.name,
-                'version': kernel_pkg.version,
-                'source_name': kernel_pkg.source_name,
-                'source_version': kernel_pkg.source_version,
-                'arch': kernel_pkg.architecture,
+                'name': kernel_pkg.name,
+                'version': kernel_pkg.installed.version,
+                'source_name': kernel_pkg.installed.source_name,
+                'source_version': kernel_pkg.installed.source_version,
+                'arch': kernel_pkg.installed.architecture
             }
     else:
         # For rpm-based distros.
         import rpm
         ts = rpm.ts()
         mi = ts.dbMatch()
-        packages = [package_header for package_header in mi if boot_image.encode() in
+        packages = [package_header for package_header in mi if boot_image_path.encode() in
                     package_header[rpm.RPMTAG_FILENAMES]]
         if packages:
             kernel_pkg = packages[0]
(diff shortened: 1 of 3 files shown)


vpetersson:
@rptrchv Check the Codacity errors. It's complaining about code quality.

rptrchv:
@a-martynovich 

> First, you use regexes for parsing the version from the package name. This will easily break because the package name may contain more dashes than your regex expects (look at https://packages.debian.org/jessie/linux-image-3.16.0-10-amd64 section "similar packages).

Those kernels (that don't match my re pattern) are **very** exotic. Besides, this re pattern can be easily adjusted to support them too: `r'(.+-\d+.\d+.\d+-)(\d+)([\.-].+)'`.

>Second, you're assuming that you can simply compare those packages by one number in their name, which will also break in many cases.

For kernels of **all** (available to me for check) platforms and distro versions it works. So this solution is not perfect (will not cover non-standard versions naming) but more than satisfactory.

> There is an easier and more robust solution. For every installed linux-image-... find its parent (the package which depends on it). Kernels which are not the latest version won't have a parent. Only the kernel of the latest version will have a parent, and that (or a level higher, if Ubuntu) is the meta-package.

We can't simply check **all** installed "linux-image-..." packages because there might be installed several different kernel **types** on a machine, e.g. "linux-image-4.15.0-XXX-aws" and "linux-image-4.15.0-XXX-generic". So we need to check for parents existence only kernel image packages of the **same** type as our running kernel (that's what my "get_latest_same_kernel_deb" function does now). Which again relies on re patterns matching, which in turn is not 100% reliable (as you pointed out). So your proposed solution has the same potential issue as mine.

a-martynovich:
@rptrchv I see your point and I can't come up with any easier solution now.
I won't ask you to write unit tests because I guess it's going to be more code than the real code itself. But please list which distros and which kernel packages you've tested on.

agent/os_helper.py
```diff
@@ -220,18 +222,13 @@ def auto_upgrades_enabled():
...
         # 2. check if it's running
         # 3. check if it has proper values in its config file
         import rpm
-        try:
-            from sh import systemctl
-        except ImportError:
-            # No systemd - probably yum-cron is not running
-            # TODO: use "service" executable which also works without systemd and on older systems
-            return False
+        from sh import systemctl
```
a-martynovich: I wrote this so it wouldn't fail if running in container. Why did you change this?

agent/os_helper.py
```diff
@@ -303,3 +316,52 @@ def apply(self, pkg):
...
+    if is_debian():  # For apt-based distros.
+        import apt_pkg
+        apt_pkg.init()
+        kernel_pkg = get_kernel_deb_package(boot_image_path)
+        if kernel_pkg is not None:
+            match = DEBIAN_KERNEL_PKG_NAME_RE.match(kernel_pkg.name)
+            if match:
+                name_parts = match.groups()  # E.g. ('linux-image-4.4.0-', '174', '-generic')
+                latest_kernel_pkg = get_latest_same_kernel_deb(name_parts[0], name_parts[2])
+                return apt_pkg.version_compare(latest_kernel_pkg.installed.version, kernel_pkg.installed.version) > 0
```
a-martynovich: why only newer? Can't we install older version? I think this should be `!= 0` i.e. not equal.

agent/os_helper.py
```diff
@@ -303,3 +316,52 @@ def apply(self, pkg):
...
+    if boot_image_path is None:
+        return None
+    if is_debian():  # For apt-based distros.
+        import apt_pkg
+        apt_pkg.init()
+        kernel_pkg = get_kernel_deb_package(boot_image_path)
+        if kernel_pkg is not None:
+            match = DEBIAN_KERNEL_PKG_NAME_RE.match(kernel_pkg.name)
+            if match:
+                name_parts = match.groups()  # E.g. ('linux-image-4.4.0-', '174', '-generic')
```
a-martynovich: How about we change this regex to `linux-image-\d+\.\d+\.\d+\.-\d+-.+` since we actually expect it to look like that? This way we don't accidentally catch packets which we don't know.

agent/os_helper.py
```diff
@@ -69,31 +71,27 @@ def is_debian():
...
+    elif is_amazon_linux2():  # For Amazon Linux 2.
+        import rpm
+        ts = rpm.ts()
+        mi = ts.dbMatch('name', 'python3-wott-agent')
+        if mi.count() > 0:
+            package_header = next(mi)
+            if __file__.encode() in package_header[rpm.RPMTAG_FILENAMES]:
+                return Installation.RPM
     # Other.
+    import agent
```
a-martynovich: `from agent import __version__`?

agent/os_helper.py
```diff
@@ -261,38 +289,23 @@ def kernel_package():
...
-                'name': kernel_pkg.package.name,
-                'version': kernel_pkg.version,
-                'source_name': kernel_pkg.source_name,
-                'source_version': kernel_pkg.source_version,
-                'arch': kernel_pkg.architecture,
+                'name': kernel_pkg.name,
+                'version': kernel_pkg.installed.version,
+                'source_name': kernel_pkg.installed.source_name,
+                'source_version': kernel_pkg.installed.source_version,
+                'arch': kernel_pkg.installed.architecture
```
a-martynovich: you don't need `.installed` because `get_kernel_deb_package` already checks for that (`is_installed`).

agent/os_helper.py
```diff
@@ -303,3 +316,52 @@ def apply(self, pkg):
...
+        return None
+    if is_debian():  # For apt-based distros.
+        import apt_pkg
+        apt_pkg.init()
+        kernel_pkg = get_kernel_deb_package(boot_image_path)
+        if kernel_pkg is not None:
+            match = DEBIAN_KERNEL_PKG_NAME_RE.match(kernel_pkg.name)
+            if match:
+                name_parts = match.groups()  # E.g. ('linux-image-4.4.0-', '174', '-generic')
+                latest_kernel_pkg = get_latest_same_kernel_deb(name_parts[0], name_parts[2])
```
a-martynovich: this may break in case of `linux-gcp` to `linux-gcp-edge` or `linux-signed-gcp-edge` upgrade. Basically it will be `linux-image-4.15.0-xxxx-gcp` -> `linux-image-4.18.0-xxxx-gcp`.

agent/os_helper.py
rptrchv: > I suggest renaming is_amazon2()

I don't agree. We can do that only after testing everything on RHEL and CentOS, not beforehand.

agent/os_helper.py
rptrchv: We need reboot only when we have a newer version of the running kernel package installed. That exactly what my code does. I don't catch your point.

agent/os_helper.py
rptrchv: `.installed` here is a quite different thing

agent/os_helper.py
rptrchv: Yes. But it's not that simple to properly handle it. That's why I believe the major kernel version upgrade case is out of the scope of this task and deserves a separate task.

agent/os_helper.py
a-martynovich: you replied to a different comment

vpetersson:
@a-martynovich can you give me a tl;dr how you think this should be redesigned? In the meantime, I'm merging this.

a-martynovich:
I think it should build a dependency tree from any known kernel meta-package down to a concrete kernel package and not rely on package names. This also has some edge cases, though.

vpetersson:
Ok, I'll create a new ticket for that.

a-martynovich:
@vpetersson we still need to expand this on Debian support, so yes, a ticket is needed.
On Debian everything is completely different, as we've recently learned.

PR #282, opened 2020-03-18
Fix systemd scripts in Amazon.

a-martynovich:
Fixes auto-update and the daemon service.
Both services get restarted and continue to work after package upgrade.

### DIFF
--- a/debian/wott-agent-self-update.sh
+++ b/debian/wott-agent-self-update.sh
@@ -3,5 +3,19 @@
 # -*- sh-basic-offset: 4 -*-
 
 set -euo pipefail
-apt-get update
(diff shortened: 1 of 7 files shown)


debian/wott-agent-self-update.sh
```diff
@@ -3,5 +3,19 @@
 # -*- sh-basic-offset: 4 -*-
 
 set -euo pipefail
-apt-get update
-apt-get install -ys wott-agent
+
+DISTRO=$(grep ^ID= /etc/os-release | sed 's/ID=//g; s/"//g')
+case $DISTRO in
```
vpetersson: Nice solution.

PR #284, opened 2020-03-20
Package upgrade command

a-martynovich:
New command: `wott-agent upgrade pkg [pkg...]`
#257

### DIFF
--- a/agent/os_helper.py
+++ b/agent/os_helper.py
@@ -388,7 +388,7 @@ def reboot_required():
 
 
 def confirmation(message):
-    yesno = input(message+" [y/N]")
+    yesno = input(message + " [y/N]")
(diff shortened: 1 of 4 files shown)


agent/os_helper.py
```diff
@@ -380,3 +380,25 @@ def reboot_required():
                                        key=cmp_to_key(rpm.versionCompare), reverse=True)[0]
             return rpm.versionCompare(latest_kernel_pkg, kernel_pkg) > 0
     return None
+
+
+def upgrade_packages(pkg_names):
```
vpetersson: Some thoughts:
* We need to log/print the packages that will be upgrade
* We need to ask for confirmation from the user before we proceed

a-martynovich: done

PR #287, opened 2020-03-23
Compatibility with older versions of python3-apt.

a-martynovich:
Fixes an error on Ubuntu Xenial and Debian Stretch.
#257

### DIFF
--- a/agent/os_helper.py
+++ b/agent/os_helper.py
@@ -406,7 +406,9 @@ def upgrade_packages(pkg_names):
         cache.update(apt.progress.text.AcquireProgress())
         cache.open()
         for pkg_name in unique_names:
-            pkg = cache.get(pkg_name)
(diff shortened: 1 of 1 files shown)


agent/os_helper.py
```diff
@@ -406,7 +406,9 @@ def upgrade_packages(pkg_names):
         cache.update(apt.progress.text.AcquireProgress())
         cache.open()
         for pkg_name in unique_names:
-            pkg = cache.get(pkg_name)
+            # Older versions of python3-apt don't provide full dict interface, namely .get().
+            # The result of this expression will either be False or a apt.package.Package instance.
+            pkg = pkg_name in cache and cache[pkg_name]
```
rptrchv: Instead of using readable `if` clause, you used a very tricky unreadable construction and had to add a special comment explaining it

